PR #7, opened 2023-02-15
Add maintenance commands

rafaelgomesxyz:
Adds:

- `maintenance` to check whether maintenance mode is on or off for an app
- `maintenance:on` to enable maintenance mode for an app
- `maintenance:off` to disable maintenance mode for an app
- `maintenance:set-page` to set the page for maintenance mode for an app (only works if maintenance workload uses `shakacode/maintenance-mode` image)

Maintenance mode is done by having a basic `maintenance_workload` that serves a static HTML page (provided through the `PAGE_URL` env var), and it only works for domains that use path based routing mode and have a route configured for the prefix `/` on either port 80 or 443.

The process of enabling maintenance mode is:

- Start `maintenance_workload`
- Wait until it's ready
- Switch workload for the domain to `maintenance_workload`
- Wait 30 seconds to give the domain time to update
- Stop all other workloads

And the process of disabling it is the reverse:

- Start all other workloads
- Wait until they're ready
- Switch workload for the domain to `one_off_workload`
- Wait 30 seconds to give the domain time to update
- Stop `maintenance_workload`

Example output for `maintenance:on`:

Code_vUSWUBSH7J

Example output for `maintenance:off`:

Code_eYGXrtalmK

Example output for `maintenance:set-page`:

Code_7V6NxzcElg

I only found one issue sometimes: when disabling maintenance mode, I kept refreshing the page to see it change from maintenance back to the app, and sometimes a refresh would lead to this:

msedge_Wienr87ehc

I couldn't pinpoint exactly why, but it might be something with Control Plane, since it happened within the 30 seconds window that we wait.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -173,7 +173,7 @@ cpl logs -a $APP_NAME -w $WORKLOAD_NAME
 - Checks if maintenance mode is on or off for an app
 - Outputs 'on' or 'off'
 - Specify the one-off workload through `one_off_workload` in the `.controlplane/controlplane.yml` file
-- Specify the maintenance workload through `maintenance_workload` in the `.controlplane/controlplane.yml` file
+- Optionally specify the maintenance workload through `maintenance_workload` in the `.controlplane/controlplane.yml` file (defaults to 'maintenance')
 - Maintenance mode is only supported for domains that use path based routing mode and have a route configured for the prefix '/' on either port 80 or 443
 
 ```sh
@@ -184,7 +184,7 @@ cpl maintenance -a $APP_NAME
 
 - Disables maintenance mode for an app
 - Specify the one-off workload through `one_off_workload` in the `.controlplane/controlplane.yml` file
-- Specify the maintenance workload through `maintenance_workload` in the `.controlplane/controlplane.yml` file
+- Optionally specify the maintenance workload through `maintenance_workload` in the `.controlplane/controlplane.yml` file (defaults to 'maintenance')
 - Maintenance mode is only supported for domains that use path based routing mode and have a route configured for the prefix '/' on either port 80 or 443
 
 ```sh
@@ -195,7 +195,7 @@ cpl maintenance:off -a $APP_NAME
 
 - Enables maintenance mode for an app
 - Specify the one-off workload through `one_off_workload` in the `.controlplane/controlplane.yml` file
-- Specify the maintenance workload through `maintenance_workload` in the `.controlplane/controlplane.yml` file
+- Optionally specify the maintenance workload through `maintenance_workload` in the `.controlplane/controlplane.yml` file (defaults to 'maintenance')
 - Maintenance mode is only supported for domains that use path based routing mode and have a route configured for the prefix '/' on either port 80 or 443
 
 ```sh
@@ -207,7 +207,7 @@ cpl maintenance:on -a $APP_NAME
 - Sets the page for maintenance mode
 - Only works if the maintenance workload uses the `shakacode/maintenance-mode` image
 - Will set the URL as an env var `PAGE_URL` on the maintenance workload
-- Specify the maintenance workload through `maintenance_workload` in the `.controlplane/controlplane.yml` file
+- Optionally specify the maintenance workload through `maintenance_workload` in the `.controlplane/controlplane.yml` file (defaults to 'maintenance')
 
 ```sh
 cpl maintenance:set-page PAGE_URL -a $APP_NAME
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -172,8 +172,9 @@ cpl logs -a $APP_NAME -w $WORKLOAD_NAME
 
 - Checks if maintenance mode is on or off for an app
 - Outputs 'on' or 'off'
+- Specify the one-off workload through `one_off_workload` in the `.controlplane/controlplane.yml` file
 - Specify the maintenance workload through `maintenance_workload` in the `.controlplane/controlplane.yml` file
-- Specify the domain through `domain` in the `.controlplane/controlplane.yml` file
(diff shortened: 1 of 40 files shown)


justin808:
What's next on this one?

rafaelgomesxyz:
Need to double check with Sergey if we're still doing this approach, because it led to some DNS issues last time he tried.

justin808:
@dzirtusss @rafaelgomesxyz let's get this one completed.

justin808:
Let’s finish up https://github.com/shakacode/heroku-to-control-plane/pull/7 ==> Supposedly, the approach should be valid. We just had a problem with apex domains

justin808:
Just spoke to @dzirtusss. We need to hold off on this for a bit.

justin808:
See https://shakacode.slack.com/archives/C03RWL8EP5Y/p1682534389198079?thread_ts=1682532903.102339&cid=C03RWL8EP5Y

justin808:
@rafaelgomesxyz, @dzirtusss says let's use a public maint image on our public dockerhub, like `shakacode/maintenance-mode` 

and then the arg is a publically accessible URL for a file, as Heroku does it.

docs/commands.md
```diff
@@ -167,6 +167,51 @@ cpl logs -a $APP_NAME
...
+
+```sh
+cpl maintenance -a $APP_NAME
+```
+
+### `maintenance:off`
+
+- Disables maintenance mode for an app
+- Specify the one-off workload through `one_off_workload` in the `.controlplane/controlplane.yml` file
+- Specify the maintenance workload through `maintenance_workload` in the `.controlplane/controlplane.yml` file
```
justin808, line 185: same as prior comment.

docs/commands.md
```diff
@@ -167,6 +167,51 @@ cpl logs -a $APP_NAME
...
+- Maintenance mode is only supported for domains that use path based routing mode and have a route configured for the prefix '/' on either port 80 or 443
+
+```sh
+cpl maintenance:off -a $APP_NAME
+```
+
+### `maintenance:on`
+
+- Enables maintenance mode for an app
+- Specify the maintenance workload through `maintenance_workload` in the `.controlplane/controlplane.yml` file
```
justin808, line 196: same

examples/controlplane.yml
```diff
@@ -23,6 +23,9 @@ aliases:
       - postgres
       - memcached
 
+    # Configure the workload name used when maintenance mode is on
+    maintenance_workload: maintenance
```
justin808, line 27: Need to mention the default one.

lib/command/maintenance.rb
```diff
@@ -0,0 +1,29 @@
...
+  class Maintenance < Base
+    NAME = "maintenance"
+    OPTIONS = [
+      app_option(required: true)
+    ].freeze
+    DESCRIPTION = "Checks if maintenance mode is on or off for an app"
+    LONG_DESCRIPTION = <<~DESC
+      - Checks if maintenance mode is on or off for an app
+      - Outputs 'on' or 'off'
+      - Specify the maintenance workload through `maintenance_workload` in the `.controlplane/controlplane.yml` file
```
justin808, line 13: default one?

lib/command/maintenance.rb
```diff
@@ -0,0 +1,29 @@
...
+      - Specify the domain through `domain` in the `.controlplane/controlplane.yml` file
+    DESC
+
+    def call
+      maintenance_workload = config[:maintenance_workload]
+      domain = config[:domain]
+
+      workload = cp.domain_get_workload(domain)
```
justin808, line 21: Most other commands work on the app name. Why not specify the production app name?

lib/command/maintenance_set_page.rb
```diff
@@ -0,0 +1,34 @@
...
+    REQUIRES_ARGS = true
+    OPTIONS = [
+      app_option(required: true)
+    ].freeze
+    DESCRIPTION = "Sets the page for maintenance mode"
+    LONG_DESCRIPTION = <<~DESC
+      - Sets the page for maintenance mode
+      - Only works if the maintenance workload uses the `shakacode/maintenance-mode` image
+      - Will set the URL as an env var `PAGE_URL` on the maintenance workload
+      - Specify the maintenance workload through `maintenance_workload` in the `.controlplane/controlplane.yml` file
```
justin808, line 16: Does this name really need to be configurable?

Why not always use `maintenance-workload`

justin808 (changes requested):
See comments

lib/command/maintenance.rb
```diff
@@ -0,0 +1,37 @@
...
+      app_option(required: true)
+    ].freeze
+    DESCRIPTION = "Checks if maintenance mode is on or off for an app"
+    LONG_DESCRIPTION = <<~DESC
+      - Checks if maintenance mode is on or off for an app
+      - Outputs 'on' or 'off'
+      - Specify the one-off workload through `one_off_workload` in the `.controlplane/controlplane.yml` file
+      - Optionally specify the maintenance workload through `maintenance_workload` in the `.controlplane/controlplane.yml` file (defaults to 'maintenance')
+      - Maintenance mode is only supported for domains that use path based routing mode and have a route configured for the prefix '/' on either port 80 or 443
+    DESC
```
justin808, line 16: Someday, we should consider a way to run the commands or parse the text to generate the commands.md file. Otherwise, these might be out of sync.

@rafaelgomesxyz can you search to see if somebody has already solved this problem?

justin808 (approved):

justin808:
@rafaelgomesxyz please rebase and merge and update CHANGELOG for all recent merges.

lib/command/maintenance.rb
rafaelgomesxyz: We have a task for that: `bundle exec rake command_docs`. It updates `commands.md` from the commands. I'll see if I can add a test to make sure that we don't forget to run it.

PR #16, opened 2023-03-12
Add release script

ahangarha:


ahangarha:
@justin808 
It would be good if you can test this. It should work.

justin808 (approved):

Issue #8, opened 2023-02-22
Updates to CLI Ideas


justin808:
# Pending Name Changes
- [ ] `promote-image` ==> `deploy-image`

# Missing Commands
- [ ] `promote-app-from-upstream -a hichee-production` ==> uses the `upstream` yaml app name. Includes steps like the below CI example

# Example of promote-app-from-upstream done via CircleCI steps

```yaml
  controlplane-release-production:
    docker:
      # this node version is for for controlplane cli, it may be different from hichee node version
      - image: cimg/ruby:3.1-node
    # resource_class: large
    environment:
      CPLN_ORG_STAGING: hichee-staging
      APP_NAME_STAGING: hichee-staging
      CPLN_ORG_PRODUCTION: hichee-production
      APP_NAME_PRODUCTION: hichee-production
    steps:
      - checkout
      - setup_remote_docker:
          docker_layer_caching: true
      - run:
          name: Install Control Plane tools
          command: |
            sudo npm install -g @controlplane/cli && cpln --version

            gem install cpl
      - run:
          name: Copy image
          command: |
            cpln profile create default --token ${CPLN_TOKEN} --org ${CPLN_ORG_STAGING} --gvc ${APP_NAME_STAGING}

            IMAGE_NAME_STAGING=$(cpl latest-image -a hichee-staging)
            IMAGE_NAME_PRODUCTION=$(echo $IMAGE_NAME_STAGING | sed "s/[^:]*/${APP_NAME_PRODUCTION}/")

            IMAGE_URL_STAGING=${CPLN_ORG_STAGING}.registry.cpln.io/${IMAGE_NAME_STAGING}
            IMAGE_URL_PRODUCTION=${CPLN_ORG_PRODUCTION}.registry.cpln.io/${IMAGE_NAME_PRODUCTION}

            echo $IMAGE_URL_STAGING
            echo $IMAGE_URL_PRODUCTION

            cpln image docker-login
            docker pull ${IMAGE_URL_STAGING}

            CPLN_TOKEN=$CPLN_TOKEN_PRODUCTION
            cpln profile update default --token ${CPLN_TOKEN} --org ${CPLN_ORG_PRODUCTION} --gvc ${APP_NAME_PRODUCTION}
            cpln image docker-login

            docker tag ${IMAGE_URL_STAGING} ${IMAGE_URL_PRODUCTION}

            docker push ${IMAGE_URL_PRODUCTION}
      - run:
[...]

PR #23, opened 2023-03-22
Add info command

rafaelgomesxyz:
Partial fix for #8

Also adds an `org` option (and the ability to specify a default org through `default_cpln_org` in `controlplane.yml`), because we need to be able to run `cpl info` without providing an app.

Example output:

Code_kg9NkJeVaV

### DIFF
--- a/lib/command/info.rb
+++ b/lib/command/info.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+module Command
+  class Info < Base
+    NAME = "info"
+    OPTIONS = [
+      org_option,
+      app_option
+    ].freeze
+    DESCRIPTION = "Displays a list of available workloads for all GVCs or a specific GVC in an org"
+    LONG_DESCRIPTION = <<~HEREDOC
+      - Displays a list of available workloads for all GVCs or a specific GVC in an org
+    HEREDOC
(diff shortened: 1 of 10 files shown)


lib/command/info.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+module Command
+  class Info < Base
+    NAME = "info"
+    OPTIONS = [
+      org_option,
+      app_option
+    ].freeze
+    DESCRIPTION = "Displays a list of available workloads for all GVCs or a specific GVC in an org"
```
justin808, line 10: We need to resolve the calling of GVC vs. app. Below, we have $APP_NAME.

For the info, we should be displaying Apps

```
DESCRIPTION = "Displays a list of available workloads for all Apps or a specific App in an org. Apps equal GVCs"
```

lib/core/controlplane.rb
```diff
@@ -55,6 +59,10 @@ def gvc_delete(a_gvc = gvc)
 
   # workload
 
+  def fetch_workloads(a_gvc = gvc)
```
justin808, line 62: @dzirtusss do you think using a method call as a default param is good?

justin808 (changes requested):

PR #24, opened 2023-03-27
Revert default org

rafaelgomesxyz:
Reverts the logic for a default org added by #23 

So the info command now:

- retrieves workloads for a single app when used with `-a` (or multiple apps if `match_if_app_name_starts_with` is `true`)
- retrieves workloads for all apps in an org when used with `-o`
- retrieves workloads for all apps in all orgs (that are defined in `.controlplane/controlplane.yml`) when used without any params

Code_94gg0sZbba

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -91,17 +91,18 @@ if [ cpl exists -a $APP_NAME ]; ...
 
 ### `info`
 
-- Displays a list of available workloads for all apps or a specific app in an org (apps equal GVCs)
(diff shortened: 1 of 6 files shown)


justin808:
@rafaelgomesxyz does this show the difference between the YML files and what's currently deployed?

rafaelgomesxyz:
> @rafaelgomesxyz does this show the difference between the YML files and what's currently deployed?

It does now after 3bdfc2171323879c52dc62881e4ef0eb9d2a7fff .

Let me know if this output looks good:

Code_vz7Md9OghQ

justin808 (approved):
LGTM

PR #25, opened 2023-03-29
Improve output

rafaelgomesxyz:
Partial fix for #8

Improves the output of all commands except for `run` and `run:detached`, because those are a bit complex, so I'll handle them separately.

| Command | Before | After |
| --- | --- | --- |
| `build-image` | build_image_before | build_image_after |
| `cleanup-old-images` | cleanup_old_images_before | cleanup_old_images_after |
| `cleanup-stale-apps` | cleanup_stale_apps_before | cleanup_stale_apps_after |
| `config` | config_before | For each app: <br/><br/> config_each_after <br/><br/> For a specific app: <br/><br/> config_specific_after |
| `delete` | delete_before | delete_after |
| `deploy-image` | deploy_image_before | deploy_image_after |
| `ps:restart` | ps_restart_before | ps_restart_after |
| `ps:start` | ps_start_before | ps_start_after |
| `ps:stop` | ps_stop_before | ps_stop_after |
| `setup` | setup_before | When the GVC doesn't exist yet: <br /><br /> setup_after <br /><br /> When the GVC already exists: <br /><br /> setup_existing_after |

Something else that would be good to improve is error handling, but I'll leave that for another time.

### DIFF
--- a/lib/core/shell.rb
+++ b/lib/core/shell.rb
@@ -18,14 +18,14 @@ class Shell
   end
 
   def self.warn(message)
-    stderr.puts(color("WARNING: #{message}\n", :yellow))
+    stderr.puts(color("WARNING: #{message}", :yellow))
   end
 
   def self.warn_deprecated(message)
-    stderr.puts(color("DEPRECATED: #{message}\n", :yellow))
+    stderr.puts(color("DEPRECATED: #{message}", :yellow))
   end
 
   def self.abort(message)
-    Kernel.abort(color("ERROR: #{message}\n", :red))
+    Kernel.abort(color("ERROR: #{message}", :red))
   end
 end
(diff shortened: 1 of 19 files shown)


lib/core/shell.rb
```diff
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
 class Shell
```
justin808, line 3: Should we have 2 class `Shell` in this project?

justin808 (changes requested):
Please rebase and merge. LGTM. 

Just a few questions.

lib/core/shell.rb
rafaelgomesxyz: You mean because of Ruby's [`Shell`](https://ruby-doc.org/stdlib-2.5.1/libdoc/shell/rdoc/Shell.html)? We could rename it, maybe to `ThorShell`?

Issue #27, opened 2023-03-29
Missing Doc Items


justin808:
- [x] Doc for how to setup cron jobs, equivalent to [Heroku Scheduler](https://elements.heroku.com/addons/scheduler)

PR #29, opened 2023-03-29
Add scheduled jobs to docs

justin808:
Partial fix for #27


justin808:
I created the example YML file.

PR #33, opened 2023-03-31
Updates to setup command

### LINKED ISSUE #28: cpl setup ==> cpl apply-template and cpl setup-app
Good point. Let's do this:

# `cpl apply-template`
1. rename `cpl setup` to `cpl apply-template`
1. Check object exists for app. If so, then require an additional option: `--confirm <app_name>` . If we can't check if the object exists, the always require `--confirm <app_name>`

# `cpl setup-app`
1. Add array of templates to optional key `setup` under the app in `controlplane.yml`

example:
```yml
apps:
  hichee-staging:
    <<: *common
  hichee-review-:
    <<: *common
    match_if_app_name_starts_with: true
    stale_app_image_deployed_days: 5
    # setup should only be used for review apps, never for persistent apps like production
    setup:
      - gvc
      - identity
      - redis
      - memcached
      - rails
      - sidekiq
```

3. Add a command that will apply 

`cpl setup-app -a <app_name>`

If app exists, return error code 1 and display message:

Error! App <app_name> already exists. If you want to update this app, either run `cpl delete -a <APP_NAME>` and rerun, or run `cpl apply-template <template1> <template2> -a <APP_NAME>`.

rafaelgomesxyz:
Fixes #28

- Ensures that all templates exist before applying
- Renames `setup` command to `apply-template`
- Checks if app/workloads exist before applying
- Adds `setup-app` command

Example outputs for `apply-template`:

- When any template is missing:

Code_ER9cSnro0J

- When app/workloads already exist:

Code_3wgJ1S5Cbo

Example outputs for `setup-app`:

- When app already exists:

Code_KyCmBtzPKw

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -11,6 +11,30 @@ This `-a` option is used in most of the commands and will pick all other app con
 
 ### Commands
 
+### `apply-template`
+
+- Applies application-specific configs from templates (e.g., for every review-app)
+- Publishes (creates or updates) those at Control Plane infrastructure
+- Picks templates from the `.controlplane/templates` directory
+- Templates are ordinary Control Plane templates but with variable preprocessing
+
+**Preprocessed template variables:**
+
+```
+APP_GVC      - basically GVC or app name
+APP_LOCATION - default location
+APP_ORG      - organization
+APP_IMAGE    - will use latest app image
+```
+
+```sh
+# Applies single template.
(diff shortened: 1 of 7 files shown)


justin808:
Right. This is tangential to #28. I'll rename that one. Yes, good to report on the invalids args ALWAYS.

justin808 (approved):

PR #40, opened 2023-05-04
Set commit as runtime env var when building image

rafaelgomesxyz:
- Sets commit as runtime env var `GIT_COMMIT` when building image, by providing it to `docker build` through `--build-arg`
- Raises error if Dockerfile can't be found when building image

### DIFF
--- a/lib/command/build_image.rb
+++ b/lib/command/build_image.rb
@@ -32,7 +32,7 @@ module Command
       build_args = []
       build_args.push("GIT_COMMIT=#{commit}") if commit
 
-      cp.image_build_with_docker(image_url, dockerfile: dockerfile, build_args: build_args)
+      cp.image_build(image_url, dockerfile: dockerfile, build_args: build_args)
 
       progress.puts("\nPushed image to '/org/#{config.org}/image/#{image_name}'.")
     end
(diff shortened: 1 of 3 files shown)


lib/command/build_image.rb
```diff
@@ -11,18 +11,30 @@ class BuildImage < Base
...
       ensure_docker_running!
 
       dockerfile = config.current[:dockerfile] || "Dockerfile"
       dockerfile = "#{config.app_cpln_dir}/#{dockerfile}"
 
+      raise "Can't find Dockerfile at '#{dockerfile}'." unless File.exist?(dockerfile)
+
       progress.puts("Building image from Dockerfile '#{dockerfile}'...\n\n")
 
-      cp.image_build(latest_image_next, dockerfile: dockerfile)
```
justin808, line 25: @rafaelgomesxyz should we remove `cp.image_build`?

justin808 (changes requested):
clean up image_build?

lib/command/build_image.rb
rafaelgomesxyz: I wasn't sure if we wanted to keep it around to maybe use it for something else, but ok, removed it and renamed `image_build_with_docker` to `image_build`.

justin808: why image_build and not build_image?

justin808 (approved):

lib/command/build_image.rb
rafaelgomesxyz: Just keeping it consistent with Sergey's original function names, which are all in the `noun + verb` form. We might want to rename all methods to the `verb + noun` form, we just have to be consistent.

PR #44, opened 2023-05-11
Updates to run commands

rafaelgomesxyz:
- Ensures that we get the terminal size from a common output on Linux and Mac
- Improves the output
- Defaults to port 80 when not set
- Checks if the workload still exists before deleting it
- Adds macOS to CI

Example outputs:

| Command | Before | After |
| --- | --- | --- |
| `run` | Code_LnPcwzVNVo | Code_HHPIN4bP4Z |
| `run:detached` | Code_tyfaJGEtKl | Code_fphx2KqrFq |

### DIFF
--- a/.github/workflows/rspec.yml
+++ b/.github/workflows/rspec.yml
@@ -8,15 +8,18 @@ on:
 
 jobs:
   rspec:
-    runs-on: ubuntu-latest
-    name: RSpec
-    env:
-      RAILS_ENV: test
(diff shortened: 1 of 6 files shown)


justin808 (approved):
LGTM. Can I get confirmation that this works on MacOS? @dzirtusss

justin808:
@rafaelgomesxyz once confirmed on MacOS, please merge.

PR #50, opened 2023-06-06
Improve docs

### LINKED ISSUE #38: Document how to install and use
Should `cpl` be in the Gemfile?

I think no (just need to lock before 1.0)

After 1.0, we should:

* Print warning if using non-latest version with message how to update.

Plus, we need to have some constant with the min version of `cpln` supported, and exit if current version is less than that with instructions on how to update this.

### LINKED ISSUE #42: Document for Sidekiq Tips
@rafaelgomesxyz Need to document these for sidekiq:

# quiet non-critical workers during deployments
https://github.com/shakacode/hichee/blob/master/.circleci/config.yml#L559-L567

```yml
      - run:
          # NOTE: doing this early enough to give time for all workers to finish jobs gracefully
          # before new image will be deployed. No need for un-quiet as this will happen automatically
          # after new image will be promoted.
          name: Quiet sidekiq workers (except criticalworker) to avoid locks in migrations
          command: |
            CPLN_TOKEN=$CPLN_TOKEN_PRODUCTION
            cpln profile create default --token ${CPLN_TOKEN} --org ${CPLN_ORG_PRODUCTION} --gvc ${APP_NAME_PRODUCTION}
            cpl run 'rails runner "Sidekiq::ProcessSet.new.each { |w| w.quiet! unless w[%q(hostname)].start_with?(%q(criticalworker.)) }"' -a ${APP_NAME_PRODUCTION}
```

# Proper setting of pre-kill hook

Guys, after testing sidekiq I've come to add a pre-stop hook to sidekiq. It works as follows:
* re-deployment is initiated from any place - UI, CLI, etc.
* the hook is run, and it sends "QUIET" to sidekiq workers
* sidekiq gracefully finishes jobs and waits several minutes before the replica is actually killed by cpln

This works very nicely, especially in big replica qty. 
 
I guess we should implement it for all sidekiq's workers as a template.

P.S. technically:
* can't use process name as regex, because it is ruby not sidekiq
* need to use space on sidekiq name, otherwise, it sends TSTP to sidekiqswarm process as well, and :man-shrugging: for whatever reason it doesn't work
* so... with ^ and \s we are mostly guaranteed to send it to only worker processes

image

### LINKED ISSUE #43: Need to transfer docs from internal Notion guide to here
We've got many tips in our ShakaCode Notion.

This issue is to remember to copy those over this repo.

rafaelgomesxyz:
- Minor updates to README.md
- Updates examples
- Updates templates
- Fixes headings in commands.md
- Adds check for cpln/cpl version (fixes #38)
- Adds Sidekiq tips (fixes #42)
- Adds more tips from internal Notion (fixes #43)


justin808:
AWESOME work @rafaelgomesxyz.

PR #55, opened 2023-06-12
Fix cleanup-stale-apps command for when app does not have image

rafaelgomesxyz:
When an app doesn't have an image, the `cleanup-stale-apps` command fails because it tries to access `image["created"]`, which doesn't exist.


justin808:
@rafaelgomesxyz need to update the changelist

PR #58, opened 2023-06-19
Add ps:wait command

### LINKED ISSUE #54: Add analogue of "wait" commands
There is `ps:wait` commands on heroku which are rather useful. IMO, should not be hard to implement some analogue to wait for all replicas with old deployment to finish running (and destroyed).

This is useful for final "smoke-check" to be sure that requests are going only to newly deployed replicas and not accidentally served by old deprovisioning replicas.

rafaelgomesxyz:
Fixes #54

This command waits until all replicas in all deployments in all workloads (or a specific workload) are ready with the next deployed version.

### DIFF
--- a/lib/core/controlplane.rb
+++ b/lib/core/controlplane.rb
@@ -149,22 +149,22 @@ class Controlplane # rubocop:disable Metrics/ClassLength
     api.workload_deployments(workload: workload, gvc: gvc, org: org)
   end
 
-  def deployment_version_ready?(version, next_version, ready:)
+  def workload_deployment_version_ready?(version, next_version, expected_status:)
     return false unless version["workload"] == next_version
 
     version["containers"]&.all? do |_, container|
-      ready_status = container.dig("resources", "replicas") == container.dig("resources", "replicasReady")
-      ready ? ready_status : !ready_status
+      ready = container.dig("resources", "replicas") == container.dig("resources", "replicasReady")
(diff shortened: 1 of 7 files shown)


lib/core/controlplane.rb
```diff
@@ -149,6 +149,26 @@ def fetch_workload_deployments(workload)
...
+  end
+
+  def wait_for_workload_deployments(workload, ready:)
+    deployments = fetch_workload_deployments(workload)["items"]
+    deployments.all? do |deployment|
+      next_version = deployment.dig("status", "expectedDeploymentVersion")
+
+      deployment.dig("status", "versions")&.all? do |version|
+        deployment_version_ready?(version, next_version, ready: ready)
+      end
```
justin808, line 168: I don't understand how this waits. Should there be some `sleep` and a loop?

rafaelgomesxyz: It doesn't. The method that waits is `step`, which calls this method:

step

Then it keeps sleeping and looping for as long as `wait_for_workload_deployments` returns `false`.

I think this was a leftover of the initial implementation I made and I forgot to rename it.

I've renamed it to `workload_deployments_ready?` in c02a5ef829a042f7fe6bc29f6e432d76975323c6, and also renamed the param to `expected_status`, which makes more sense.

justin808:
@rafaelgomesxyz looks amazing!

PR #83, opened 2023-10-19
Small improvements

### LINKED ISSUE #82: Small Fixes Needed
- [x] Command to open the app (gvc) on control plane web
- [x] Global `--verbose` (maybe `-d`) flag to echo commands and all output, both regular and error
- [x] `cpl` with no subcommand defaults to help.
- [x] Allow cpln_org in control_plan.yml to be blank and have ENV `CPLN_ORG` used. That way, tutorial users don't need to modify the controlplane.yml file.
- [x] Have `cpl config` indicate if the org comes from ENV value.
- [ ] Have some default troubleshooting tips, such check your value for the org...and suggest putting in an ENV. We'll have lots of example apps, so this will be a big help.

```
    cpln_org: mostafa-demo
```

rafaelgomesxyz:
Fixes #82

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,16 @@ _Please add entries here for your pull requests that are not yet released._
 
 - Fixed failed build on MacOS by adding platform flag and fixed multiple files in yaml document for template. [PR 81](https://github.com/shakacode/heroku-to-control-plane/pull/81) by [justin808](https://github.com/justin808).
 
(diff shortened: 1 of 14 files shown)


justin808:
@rafaelgomesxyz could you rebase?

justin808 (changes requested):
Two tiny issues. Can be done in follow up PR.

PR #92, opened 2023-10-27
Small fixes

justin808:


README.md
```diff
@@ -153,18 +153,31 @@ Here's a complete example of all supported config keys explained for the `contro
...
+    # Control Plane offers the ability to use multiple locations.
+    # default_location is used for commands that require a location:
+    # including `ps`, `run`, `run:detached`, `apply-template`
+    # This can be overridden with option --location=<location>
     default_location: aws-us-east-2
 
     # Allows running the command `cpl setup-app`
     # instead of `cpl apply-template gvc redis postgres memcached rails sidekiq`.
-    setup:
+    setup_app_templates_workloads:
```
ahangarha, line 180: Isn't it better to have backward compatibility?
Since we will implement a file validation process in the project, we can show a deprecation message and remove this in the next major release.

ahangarha: And why just have an entry for `app_workloads` and not `additional_workloads`?

If we want to have a way to apply all templates (especially the first time), why not accept `*` wildcard in `cpl apply_template *`? And maybe we can have special names like `_app_workloads_` to make it easier to apply only a specific category of workloads.

rafaelgomesxyz:
Closing as the stuff in here is being added by other separate PRs.

PR #94, opened 2023-11-02
Show org and app on every command

ahangarha:
A PR for showing ORG and APP in every command except certain ones.

Sample:
```
~ CPLN_ORG=my-org-name CPLN_APP=tutorial-app cpl apply-template redis
ORG: my-org-name
APP: tutorial-app
...
```

### DIFF
--- a/lib/command/base.rb
+++ b/lib/command/base.rb
@@ -23,7 +23,7 @@ module Command
     EXAMPLES = ""
     # If `true`, hides the command from `cpl help`
     HIDE = false
-    # weather to show key information like ORG and APP name in commands
+    # Whether or not to show key information like ORG and APP name in commands
     WITH_INFO_HEADER = true
 
     NO_IMAGE_AVAILABLE = "NO_IMAGE_AVAILABLE"
--- a/lib/command/base.rb
+++ b/lib/command/base.rb
@@ -28,6 +28,20 @@ module Command
 
     def initialize(config)
       @config = config
+
+      puts_info_header
+    end
+
+    def puts_info_header
+      rows = {}
+      rows["ORG"] = config.org unless config.org.nil? || config.org.empty?
+      rows["APP"] = config.org unless config.app.nil? || config.app.empty?
+
+      max_key_length = rows.keys.map(&:size).max
+
+      rows.each do |key, value|
+        puts "#{key.ljust(max_key_length)}: #{value}"
+      end
     end
 
     def self.all_commands
--- a/lib/command/base.rb
+++ b/lib/command/base.rb
@@ -23,25 +23,13 @@ module Command
     EXAMPLES = ""
     # If `true`, hides the command from `cpl help`
     HIDE = false
+    # weather to show key information like ORG and APP name in commands
+    WITH_INFO_HEADER = true
 
     NO_IMAGE_AVAILABLE = "NO_IMAGE_AVAILABLE"
 
     def initialize(config)
       @config = config
-
-      puts_info_header
-    end
-
-    def puts_info_header
-      rows = {}
-      rows["ORG"] = config.org unless config.org.nil? || config.org.empty?
-      rows["APP"] = config.org unless config.app.nil? || config.app.empty?
-
-      max_key_length = rows.keys.map(&:size).max
-
-      rows.each do |key, value|
-        puts "#{key.ljust(max_key_length)}: #{value}"
-      end
     end
 
     def self.all_commands
--- a/lib/command/info.rb
+++ b/lib/command/info.rb
@@ -27,6 +27,8 @@ module Command
       ```
     EX
 
+    WITH_INFO_HEADER = false
+
     def call
       @missing_apps_workloads = {}
       @missing_apps_starting_with = {}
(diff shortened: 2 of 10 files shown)


ahangarha:
Should we separate this "info header" from other logs using some sort of divider?

lib/command/base.rb
```diff
@@ -23,6 +23,8 @@ class Base # rubocop:disable Metrics/ClassLength
     EXAMPLES = ""
     # If `true`, hides the command from `cpl help`
     HIDE = false
+    # weather to show key information like ORG and APP name in commands
```
rafaelgomesxyz, line 26:
```suggestion
    # Whether or not to show key information like ORG and APP name in commands
```

lib/command/info.rb
```diff
@@ -27,6 +27,8 @@ class Info < Base # rubocop:disable Metrics/ClassLength
       ```
     EX
 
+    WITH_INFO_HEADER = false
```
rafaelgomesxyz, line 30: Nit: I don't think we need to separate it from the others with a newline. This applies to the other commands as well.

lib/cpl.rb
```diff
@@ -186,6 +189,18 @@ def self.all_base_commands
     rescue StandardError => e
       ::Shell.abort("Unable to load command: #{e.message}")
     end
+
+    private
+
+    def puts_info_header(config)
```
rafaelgomesxyz, line 195: Maybe `show_info_header` would be a better name?

lib/cpl.rb
```diff
@@ -186,6 +189,18 @@ def self.all_base_commands
...
+      rows = {}
+      rows["ORG"] = config.org unless config.org.nil? || config.org.empty?
+      rows["APP"] = config.app unless config.app.nil? || config.app.empty?
```
rafaelgomesxyz, line 198: Do we need to check for `nil` and `empty` or would `if config.org` and `if config.app` work?

lib/cpl.rb
```diff
@@ -186,6 +189,18 @@ def self.all_base_commands
...
+
+      rows.each do |key, value|
+        puts "#{key}: #{value}"
+      end
+    end
```
rafaelgomesxyz, line 203: I think we should put a newline at the end to separate the header from the rest of the command output.

rafaelgomesxyz (changes requested):
We should not add it to the `env` and `latest-image` commands, as their output might be used directly.

And we also need to consider the `ps` command, as it outputs a list of replicas, so if the output is used directly and we add a header it could break things. Thoughts?

lib/cpl.rb
ahangarha: What if we get an empty string for `config.app`? It is a truthy value.
I think in such cases, it is important to have a strict condition.
I am even thinking of stripping the value. Something like this:

```ruby
rows["ORG"] = config.org.to_s.strip unless config.org.to_s.strip.empty?
```

ahangarha: Also, I am wondering what is the right behavior if no org or app is provided. Currently, I don't show them. But shouldn't we put some message like `no <KEY> found!`?

rafaelgomesxyz: If we want to strip it, it might be best to do that when assigning the app/org in the `Config` class, and then we can make sure that it's always going to be either a string or `nil`. That way we can standardize how they are accessed across the entire code.

And I don't know about adding a "not found" message, because in some commands, the org/app is optional, so it might confuse users to see "not found", leading them to believe they should have specified it. Maybe it would be best to use "not provided"?

```
ORG: not provided
APP: not provided
```

ahangarha: If the user doesn't provide either of `org` or `app`, we get error messages before reaching to the point of showing the info header:

```
➜  CPLN_ORG="   my-org  " cpl deploy-image 
No value provided for required options '--app'
➜  CPLN_APP=tutorial-app cpl deploy-image 
ERROR: Can't find option 'cpln_org' for app 'tutorial-app' in 'controlplane.yml', and CPLN_ORG env var is not set. The org can also be provided through --org.
```

But the issue is with the following scenario:
```
➜ CPLN_ORG="    " CPLN_APP=tutorial-app cpl deploy-image
ORG: 
APP: tutorial-app
```

It is beyond the scope of this PR to fix this issue.
We should have a proper mechanism for all configurations.

What do you think?

lib/core/config.rb
```diff
@@ -11,9 +11,9 @@ class Config # rubocop:disable Metrics/ClassLength
   def initialize(args, options)
     @args = args
     @options = options
-    @org = options[:org]
+    @org = options[:org]&.strip
```
rafaelgomesxyz, line 14: Just be aware that we also assign `@org` in line 84 when it comes from `controlplane.yml`:

```
@org = current.fetch(:cpln_org) if current.key?(:cpln_org)
```

I don't know if `current.key?` would account for an empty string.

Just check that and strip if needed. Then feel free to merge, no need to re-request review.

rafaelgomesxyz (approved):

lib/core/config.rb
ahangarha: Good catch

PR #105, opened 2023-11-12
Accept --location option

### LINKED ISSUE #100: Accept location as an option
From #91 
---

Allow specifying `--location` for commands like `ps`, `run`, `run:detached`, and `apply-template`.

ahangarha:
This PR add a `-l, --location` option to the project, which can be used in the following commands:
- `apply-template`
- `ps`
- `run`
- `run:detached`

~Question: Should we have `CPLN_LOCATION` as well?~ --> Added in ad26c1e
Question: Should we add an option to prevent override by env variable? (like org and app)

---
Closes #100

### DIFF
--- a/lib/command/apply_template.rb
+++ b/lib/command/apply_template.rb
@@ -7,6 +7,7 @@ module Command
     REQUIRES_ARGS = true
     OPTIONS = [
       app_option(required: true),
+      location_option,
       skip_confirm_option
     ].freeze
     DESCRIPTION = "Applies application-specific configs from templates"
@@ -126,7 +127,7 @@ module Command
     def apply_template(filename)
       data = File.read(filename)
                  .gsub("APP_GVC", config.app)
-                 .gsub("APP_LOCATION", config[:default_location])
+                 .gsub("APP_LOCATION", config.options[:location] || config[:default_location])
                  .gsub("APP_ORG", config.org)
                  .gsub("APP_IMAGE", latest_image)
 
--- a/lib/command/apply_template.rb
+++ b/lib/command/apply_template.rb
@@ -127,7 +127,7 @@ module Command
     def apply_template(filename)
       data = File.read(filename)
                  .gsub("APP_GVC", config.app)
-                 .gsub("APP_LOCATION", config.options[:location] || config[:default_location])
+                 .gsub("APP_LOCATION", config.location)
                  .gsub("APP_ORG", config.org)
                  .gsub("APP_IMAGE", latest_image)
 
--- a/lib/command/ps.rb
+++ b/lib/command/ps.rb
@@ -5,6 +5,7 @@ module Command
     NAME = "ps"
     OPTIONS = [
       app_option(required: true),
+      location_option,
       workload_option
     ].freeze
     DESCRIPTION = "Shows running replicas in app"
@@ -25,12 +26,14 @@ module Command
     def call
       cp.fetch_gvc!
 
+      location = config.options["location"] || config[:default_location]
+
       workloads = [config.options[:workload]] if config.options[:workload]
       workloads ||= config[:app_workloads] + config[:additional_workloads]
       workloads.each do |workload|
         cp.fetch_workload!(workload)
 
-        result = cp.workload_get_replicas(workload, location: config[:default_location])
+        result = cp.workload_get_replicas(workload, location: location)
         result["items"].each { |replica| puts replica }
       end
     end
--- a/lib/command/ps.rb
+++ b/lib/command/ps.rb
@@ -26,7 +26,7 @@ module Command
     def call
       cp.fetch_gvc!
 
-      location = config.options["location"] || config[:default_location]
+      location = config.location
 
       workloads = [config.options[:workload]] if config.options[:workload]
       workloads ||= config[:app_workloads] + config[:additional_workloads]
--- a/lib/command/run.rb
+++ b/lib/command/run.rb
@@ -57,7 +57,7 @@ module Command
     attr_reader :location, :workload, :one_off, :container
 
     def call # rubocop:disable Metrics/MethodLength
-      @location = config.options["location"] || config[:default_location]
+      @location = config.location
       @workload = config.options["workload"] || config[:one_off_workload]
       @one_off = "#{workload}-run-#{rand(1000..9999)}"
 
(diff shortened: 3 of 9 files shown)


lib/command/apply_template.rb
```diff
@@ -126,7 +127,7 @@ def confirm_workload(template)
     def apply_template(filename)
       data = File.read(filename)
                  .gsub("APP_GVC", config.app)
-                 .gsub("APP_LOCATION", config[:default_location])
+                 .gsub("APP_LOCATION", config.options[:location] || config[:default_location])
```
rafaelgomesxyz, line 130: We could do it like `org_option` and `app_option`, and add a `default` key with the value `config[:default_location]`. See:

```
def self.org_option(required: false) # rubocop:disable Metrics/MethodLength
  {
    name: :org,
    params: {
      aliases: ["-o"],
      banner: "ORG_NAME",
      desc: "Organization name",
      type: :string,
      required: required,
      default: ENV.fetch("CPLN_ORG", nil) # <----------
    }
  }
end
```

lib/command/ps.rb
```diff
@@ -24,12 +25,14 @@ class Ps < Base
     def call
       cp.fetch_gvc!
 
+      location = config.options["location"] || config[:default_location]
```
rafaelgomesxyz, line 28: Same here.

lib/command/run.rb
```diff
@@ -56,7 +57,7 @@ class Run < Base
     attr_reader :location, :workload, :one_off, :container
 
     def call # rubocop:disable Metrics/MethodLength
-      @location = config[:default_location]
+      @location = config.options["location"] || config[:default_location]
```
rafaelgomesxyz, line 60: Same here.

lib/command/run_detached.rb
```diff
@@ -47,7 +48,7 @@ class RunDetached < Base # rubocop:disable Metrics/ClassLength
     attr_reader :location, :workload, :one_off, :container
 
     def call # rubocop:disable Metrics/MethodLength
-      @location = config[:default_location]
+      @location = config.options["location"] || config[:default_location]
```
rafaelgomesxyz, line 51: Same here.

rafaelgomesxyz (changes requested):
Just one small thing, but need to change it in multiple places.

lib/command/apply_template.rb
ahangarha: It seems at this stage, we still don't have `config`.

ahangarha:
Known issue:

Passing non-exisiting location causes error:

```
➜  cpl ps --location aws-eu-central-1
rails-7975c5cb68-vftmk
redis-5d9554cd95-czj9b
postgres-0

➜  cpl ps --location aws-us-east-1   
- message: >-
    /org/mostafa-demo/gvc/tutorial-app/workload/rails/deployment/aws-us-east-1
    does not exist
  id: b31c6c3d-3b4c-4277-b73d-1e13cf17817a
  status: 404
```

lib/command/apply_template.rb
rafaelgomesxyz: Hm, you're right, this is a class method, so it doesn't get initialized with `config`.

In that case, we can do this when initializing the `Config` class (after loading the app config and apps). Something like:

```
@location = options[:location] || config[:default_location]
```

And then have `location` in `attr_reader` so we can access it like `config.location` elsewhere.

ahangarha: I thought about it earlier. I wasn't sure about it because I am not sure of the criteria for adding methods to the `config`.

If you think this is an appropriate thing to do, I can add it.

rafaelgomesxyz: `config` is there though, you just have to do it after `load_app_config`, which is what loads `config`.

rafaelgomesxyz:
Let's also add Justin's comment from https://github.com/shakacode/heroku-to-control-plane/pull/92/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R172 here, since this PR is adding the option.

lib/core/config.rb
```diff
@@ -18,6 +18,8 @@ def initialize(args, options)
     load_app_config
     load_apps
 
+    @location = options[:location] || config.dig(:apps, app.to_sym, :default_location)
```
ahangarha, line 21: @rafaelgomesxyz This is how I could implement it.

lib/core/config.rb
```diff
@@ -18,6 +18,8 @@ def initialize(args, options)
     load_app_config
     load_apps
 
+    @location = options[:location] || config.dig(:apps, app&.to_sym, :default_location)
```
rafaelgomesxyz, line 21: We should use `config.current&.dig(:default_location)` instead. For apps where `match_if_app_name_starts_with` is `true`, `config.dig(:apps, app&.to_sym, :default_location)` wouldn't work, because the `app` provided wouldn't match the app in `controlplane.yml`.

rafaelgomesxyz (changes requested):

lib/core/config.rb
ahangarha: Oh! I see
Thanks for the hint.
I will update the code.

ahangarha:
@rafaelgomesxyz Should we check with the ControlPlane actual deployment to ensure the location we use matches the location of the deployed project?

I guess there should be multiple edge cases. We may need to discuss or address those situations in later PRs.

What do you think?

---
Notice: This PR's implementation may be affected by #110 implementation.

rafaelgomesxyz (approved):
> @rafaelgomesxyz Should we check with the ControlPlane actual deployment to ensure the location we use matches the location of the deployed project?

Maybe in another PR, I don't think it fits into the scope of this PR.

lib/core/config.rb
```diff
@@ -230,6 +234,20 @@ def load_org_from_file
...
 
+  def load_location_from_options
+    strip_str_and_validate(options[:location])
+  end
+
+  def load_location_from_env
+    strip_str_and_validate(ENV.fetch("CPLN_LOCATION", nil))
+  end
+
+  def load_location_from_file
```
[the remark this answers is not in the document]
ahangarha: Fixed. There, I noticed another mistake. That is also fixed.

Not only tested on cpln, but I also printed the stages of loading location to ensure all three scenarios run correctly.

rafaelgomesxyz (approved):

PR #109, opened 2023-11-17
Only use CPLN_ORG and CPLN_APP env vars if allowed

### LINKED ISSUE #90: Changes to how we use CPLN_ORG and CPLN_APP env vars
Should we allow these env vars only behind a setting in `controlplane.yml`?

Or add confirmation to more dangerous commands, like `deploy-image`, etc.?

Or not allow env vars at all, but suggest shortcut scripts to prevent having to provide the org/app every time?

rafaelgomesxyz:
On top of #110 for now!

Fixes  #90

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,10 +159,19 @@ Here's a complete example of all supported config keys explained for the `contro
 ```yaml
 # Keys beginning with "cpln_" correspond to your settings in Control Plane.
 
+# Global settings that apply to `cpl` usage.
+# You can opt out of allowing the use of CPLN_ORG and CPLN_APP env vars
+# to avoid any accidents with the wrong org / app.
+allow_org_override_by_env: true
+allow_app_override_by_env: true
+
 aliases:
   common: &common
-    # Organization name for staging (customize to your needs).
-    # Production apps will use a different organization, specified below, for security.
+    # Organization for staging and QA apps is typically set as an alias.
+    # Production apps will use a different organization, specified in `apps`, for security.
+    # Change this value to your organization name
+    # or set the CPLN_ORG env var and it will override this for all `cpl` commands
+    # (provided that `allow_org_override_by_env` is set to `true`).
     cpln_org: my-org-staging
 
     # Example apps use only one location. Control Plane offers the ability to use multiple locations.
@@ -229,6 +238,11 @@ apps:
 
   my-app-production:
     <<: *common
+
+    # You can also opt out of allowing the use of CPLN_ORG and CPLN_APP env vars per app.
+    # It's recommended to leave this off for production, to avoid any accidents.
+    allow_org_override_by_env: false
+    allow_app_override_by_env: false
 
     # Use a different organization for production.
     cpln_org: my-org-production
--- a/examples/controlplane.yml
+++ b/examples/controlplane.yml
@@ -1,14 +1,32 @@
 # Keys beginning with "cpln_" correspond to your settings in Control Plane.
 
+# Global settings that apply to `cpl` usage.
+# You can opt out of allowing the use of CPLN_ORG and CPLN_APP env vars
+# to avoid any accidents with the wrong org / app.
+allow_org_override_by_env: true
+allow_app_override_by_env: true
+
 aliases:
   common: &common
-    # Organization name for staging (customize to your needs).
-    # Production apps will use a different organization, specified below, for security.
+    # Organization for staging and QA apps is typically set as an alias.
+    # Production apps will use a different organization, specified in `apps`, for security.
+    # Change this value to your organization name
+    # or set the CPLN_ORG env var and it will override this for all `cpl` commands
+    # (provided that `allow_org_override_by_env` is set to `true`).
     cpln_org: my-org-staging
 
     # Example apps use only one location. Control Plane offers the ability to use multiple locations.
-    # TODO: Allow specification of multiple locations.
     default_location: aws-us-east-2
+
+    # Allows running the command `cpl setup-app`
+    # instead of `cpl apply-template gvc redis postgres memcached rails sidekiq`.
+    setup:
+      - gvc
+      - redis
+      - postgres
+      - memcached
+      - rails
+      - sidekiq
 
     # Configure the workload name used as a template for one-off scripts, like a Heroku one-off dyno.
     one_off_workload: rails
@@ -24,26 +42,62 @@ aliases:
       - postgres
       - memcached
 
-    # Configure the workload name used when maintenance mode is on (defaults to "maintenance")
+    # Configure the workload name used when maintenance mode is on (defaults to "maintenance").
     maintenance_workload: maintenance
+
+    # Fixes the remote terminal size to match the local terminal size
+    # when running the commands `cpl run` or `cpl run:detached`.
+    fix_terminal_size: true
+
+    # Apps with a deployed image created before this amount of days will be listed for deletion
+    # when running the command `cpl cleanup-stale-apps`.
+    stale_app_image_deployed_days: 5
+
+    # Images that exceed this quantity will be listed for deletion
+    # when running the command `cpl cleanup-images`.
+    image_retention_max_qty: 20
+
+    # Images created before this amount of days will be listed for deletion
+    # when running the command `cpl cleanup-images` (`image_retention_max_qty` takes precedence).
+    image_retention_days: 5
+
+    # Run workloads created before this amount of days will be listed for deletion
+    # when running the command `cpl run:cleanup`.
+    stale_run_workload_created_days: 2
 
 apps:
   my-app-staging:
     # Use the values from the common section above.
     <<: *common
+
   my-app-review:
     <<: *common
+
     # If `match_if_app_name_starts_with` is `true`, then use this config for app names starting with this name,
     # e.g., "my-app-review-pr123", "my-app-review-anything-goes", etc.
     match_if_app_name_starts_with: true
+
   my-app-production:
     <<: *common
+
+    # You can also opt out of allowing the use of CPLN_ORG and CPLN_APP env vars per app.
+    # It's recommended to leave this off for production, to avoid any accidents.
+    allow_org_override_by_env: false
+    allow_app_override_by_env: false
+
     # Use a different organization for production.
     cpln_org: my-org-production
+
     # Allows running the command `cpl promote-app-from-upstream -a my-app-production`
     # to promote the staging app to production.
     upstream: my-app-staging
+
+    # Used by the command `cpl promote-app-from-upstream` to run a release script before deploying.
+    # This is relative to the `.controlplane/` directory.
+    release_script: release_script
+
   my-app-other:
     <<: *common
+
     # You can specify a different `Dockerfile` relative to the `.controlplane/` directory (defaults to "Dockerfile").
     dockerfile: ../some_other/Dockerfile
--- a/lib/command/config.rb
+++ b/lib/command/config.rb
@@ -21,11 +21,6 @@ module Command
     EX
 
     def call # rubocop:disable Metrics/MethodLength
-      if config.org_comes_from_env
-        puts Shell.color("Org comes from CPLN_ORG env var", :red)
-        puts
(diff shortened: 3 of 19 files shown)


lib/core/config.rb
```diff
@@ -71,6 +76,57 @@ def ensure_config_app!(app_name, app_options)
...
+  def ensure_required_options!
+    missing_str = required_options
+                  .reject { |option_name| %i[org app].include?(option_name) || options.key?(option_name) }
+                  .map { |option_name| "--#{option_name}" }
+                  .join(", ")
+
+    raise "Required options missing: #{missing_str}" unless missing_str.empty?
+  end
+
+  def non_empty_str_or_nil(str)
```
rafaelgomesxyz, line 102: To make sure that `config.org` and `config.app` will always be either `nil` or a non-empty string, never an empty string.

lib/cpl.rb
```diff
@@ -157,8 +157,13 @@ def self.all_base_commands
       desc(usage, description, hide: hide)
       long_desc(long_description)
 
+      required_options = []
       command_options.each do |option|
-        method_option(option[:name], **option[:params])
+        # We'll handle required options manually in `Config`
+        required_options.push(option[:name]) if option[:params][:required]
+
+        params = option[:params].reject { |key, _| key == :required }
```
rafaelgomesxyz, line 165: We remove `required` from the params list, because we handle it manually in `Config`.

lib/core/config.rb
```diff
@@ -2,22 +2,27 @@
...
   attr_reader :config, :current,
-              :org, :org_comes_from_env, :app, :apps, :app_dir,
+              :org, :org_comes_from_env, :app, :apps, :app_dir, :required_options,
               # command line options
               :args, :options
 
   CONFIG_FILE_LOCATIION = ".controlplane/controlplane.yml"
 
-  def initialize(args, options)
+  def initialize(args, options, required_options)
```
ahangarha, line 11: Isn't it better we handle this in commands?
Commands should care about the options they must receive.

lib/cpl.rb
ahangarha: I think this is over burdening `config`.

lib/core/config.rb
rafaelgomesxyz: We're already handling `args` and `options` here, so it makes sense to handle `required_options` here too.

ahangarha: This way, we make it more difficult to change in the future.
I think we should handle this issue within each command.
The command should be responsible for checking if it gets what it is dependent on.

I am just sharing my thoughts. You have more knowledge of OOP.

rafaelgomesxyz: Handling this within each command will add a lot of verbose and repeated code.

I don't think the command should be responsible for checking if it gets what it depends on, it should only do what it needs to do and assume that all args/options will be there and will be type-validated (that's the whole point of using Thor). Anything more specific than that, sure, the command should handle it.

If we were to move this to each command we'd essentially have this in every single command:

```
unless options[:app] do
  raise "..."
end
```

Why not do it in a single place? It actually makes it easier to change it in the future.

lib/cpl.rb
rafaelgomesxyz: See: https://github.com/shakacode/heroku-to-control-plane/pull/109#discussion_r1397785197

lib/core/config.rb
rafaelgomesxyz: Hm, I don't know, I usually prefer being very explicit about method names, because then we don't even need to go to the method to know what it does / outputs. `non_empty_str_or_nil` tells me I'll get either a non-empty string or `nil`, while `sanitize` doesn't tell me anything about what it's doing.

But doing `strip` inside the method sounds like a good idea, since we're already doing that every time we call it anyway.

README.md
```diff
@@ -159,10 +159,19 @@ Here's a complete example of all supported config keys explained for the `contro
 ```yaml
 # Keys beginning with "cpln_" correspond to your settings in Control Plane.
 
+# Global settings that apply to `cpl` usage.
+# You can opt out of allowing the use of CPLN_ORG and CPLN_APP env vars
+# to avoid any accidents with the wrong org / app.
+allow_org_override_by_env: true
+allow_app_override_by_env: true
```
justin808, line 166: 1. allow opt out by app
2. default for production is to opt out in example, and default is off if not set
3. indicate that it's advisable for production to have this off

examples/controlplane.yml
```diff
@@ -1,5 +1,11 @@
 # Keys beginning with "cpln_" correspond to your settings in Control Plane.
 
+# Global settings that apply to `cpl` usage.
+# You can opt out of allowing the use of CPLN_ORG and CPLN_APP env vars
+# to avoid any accidents with the wrong org / app.
+allow_org_override_by_env: true
+allow_app_override_by_env: true
```
justin808, line 7: see prior comment

lib/core/config.rb
justin808, line 102: Should be static. Might be in a helpers class. Or here is fine, b/c this config is where we process? but if we need both for cmd args and yml, then maybe a help class?

lib/core/config.rb
```diff
@@ -81,7 +137,7 @@ def pick_current_config(app_name, app_options)
```
justin808, line 137: rename @current to @current_config

lib/core/config.rb
```diff
@@ -81,7 +137,7 @@ def pick_current_config(app_name, app_options)
 
     return if @org
 
-    @org = current.fetch(:cpln_org)&.strip if current.key?(:cpln_org)
+    @org = non_empty_str_or_nil(current.fetch(:cpln_org)&.strip) if current.key?(:cpln_org)
```
justin808, line 140: should be stripped in the `non_empty_str_or_nil`

justin808 (changes requested):

lib/core/config.rb
rafaelgomesxyz: Renamed it to `strip_str_and_validate`, what do you think about this name?

lib/command/build_image.rb
```diff
@@ -18,7 +18,7 @@ class BuildImage < Base
     def call # rubocop:disable Metrics/MethodLength
       ensure_docker_running!
 
-      dockerfile = config.current[:dockerfile] || "Dockerfile"
+      dockerfile = config.current(:dockerfile) || "Dockerfile"
```
justin808, line 21: @rafaelgomesxyz how is this. change related?

lib/core/config.rb
```diff
@@ -1,32 +1,59 @@
...
+
+    return @global_config unless key
+
+    raise "Can't find config key '#{key}' in 'controlplane.yml'." unless @global_config.key?(key)
+
+    @global_config[key]
+  end
+
+  def current(key = nil)
+    key ? @current_config&.dig(key) : @current_config
```
justin808, line 46: 1. why not just make this lazy and always take a param of the key
2. then current_config is a regular method that will throw if the config can't be parsed. 
3. Given `current_config` can't do anything if the config doesn't exist, we probably don't need to name as `current_config!`

I hate methods that have 2 types of return values based on the argument!

@dzirtusss agree?

lib/core/config.rb
```diff
@@ -1,32 +1,59 @@
...
     Shell.verbose_mode(options[:verbose])
   end
 
-  def [](key)
+  def global(key = nil)
+    key ? @global_config&.dig(key) : @global_config
+  end
+
+  def global!(key = nil)
+    ensure_global_config!
```
justin808, line 36: make this lazy per coment below.

lib/core/config.rb
```diff
@@ -1,32 +1,59 @@
...
     @args = args
     @options = options
-    @org = options[:org]&.strip
-    @org_comes_from_env = true if ENV.fetch("CPLN_ORG", nil)
-    @app = options[:app]&.strip
+    @required_options = required_options
 
-    load_app_config
+    # The order here is important
+    load_config_file
```
justin808, line 19: make this lazy?

lib/core/config.rb
```diff
@@ -1,32 +1,59 @@
...
     load_apps
+    load_app
+    load_org
```
justin808, line 22: all of these can be lazy

lib/core/config.rb
```diff
@@ -43,106 +70,176 @@ def should_app_start_with?(app_name)
...
 
-    return if @org
+    loop do
+      config_file = path + CONFIG_FILE_LOCATION
+      break config_file if File.file?(config_file)
 
-    @org = current.fetch(:cpln_org)&.strip if current.key?(:cpln_org)
-    ensure_current_config_org!(app_name)
+      path = path.parent
+      if path.root?
```
justin808, line 151: Please explain this look up logi and how it relates to the instructions.

lib/command/config.rb
```diff
@@ -21,11 +21,6 @@ class Config < Base
     EX
 
     def call # rubocop:disable Metrics/MethodLength
-      if config.org_comes_from_env
```
justin808, line 24: this file needs to change so that `load_app_config` is called lazily

justin808 (changes requested):
Just make things lazy!

No need to change all the calls!

Just change the `config.current` method to lazily load config from yaml and same for for the `config.[]` methods. 

then this PR won't have a zillion files changing as well.

lib/core/config.rb
ahangarha: What about `@current_app_config`?

rafaelgomesxyz:
@justin808 I've rebased the branch to include only changes necessary for this PR. Let's handle everything else in a separate PR.

README.md
ahangarha: Should we by default allow override or prevent it? This setup means we prevent it by default unless the user explicitly sets this option to `true`.

I think it should be opposite. We should allow by default and prevent it if it is explicitly declared.

I would suggest replacing `allow_x_override_by_env: true` to `prevent_x_overriden_by_env: false`. This way, we only prevent the override if such entry exist in the config file and it is set to `true`.

lib/core/config.rb
rafaelgomesxyz: Let's handle this in a separate PR.

lib/core/config.rb
rafaelgomesxyz: Let's handle this in a separate PR.

lib/core/config.rb
rafaelgomesxyz: Let's handle this in a separate PR.

lib/core/config.rb
rafaelgomesxyz: Let's handle this in a separate PR.

lib/core/config.rb
rafaelgomesxyz: Let's handle this in a separate PR.

lib/command/config.rb
rafaelgomesxyz: Let's handle this in a separate PR.

lib/core/config.rb
rafaelgomesxyz: Let's handle this in a separate PR.

lib/command/build_image.rb
rafaelgomesxyz: Let's handle this in a separate PR.

ahangarha:
@rafaelgomesxyz This PR might have conflicts with #110. We need to be more careful for merging.

lib/core/config.rb
```diff
@@ -137,6 +198,63 @@ def new_option_keys
...
+    allowed_locally = key_exists && current[:allow_org_override_by_env]
+    allowed_globally = !key_exists && config[:allow_org_override_by_env]
+    return unless allowed_locally || allowed_globally
+
+    @org = org_from_env
+    @org_comes_from_env = true
+  end
+
+  def load_org
+    return if @org
```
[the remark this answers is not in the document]
ahangarha: This is kind of implemented in another PR: #110 

We need to finalize that and rebase this PR on top of it.

PR #110, opened 2023-11-19
Don't read config files for some commands

ahangarha:
This PR enables us to specify commands that don't need complete configuration from config files. With this, not only can we run commands like `cpl version`, `cpl --version`, and `cpl --help`, we can add new commands like `cpl generate` to create basic file structure and config templates in a project where no Control Plane configurations exist.

The implementation is not ideal, but it is good enough to be shared and discussed.


lib/core/minimal_config.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class MinimalConfig
```
rafaelgomesxyz, line 3: Why not edit `Config` instead? We could add a named param `minimal` (or whatever) to `initialize` (that defaults to `false`) and exit before `load_app_config` if it's `true`.

That way we have a single source of truth for config, minimal or not.

rafaelgomesxyz (changes requested):

lib/core/minimal_config.rb
ahangarha: I wasn't sure if that would be a good design.
Perhaps a class that could behave differently based on a flag should become two classes.
However, this implementation also is not ideal for me.

Parallelly, I am working on a different way of handling configuration. I think it is cleaner and more flexible. But I will propose it when I make the whole system work together.

rafaelgomesxyz: Ok, maybe a better implementation:

- Update `MinimalConfig` to be exactly like `Config` before `load_app_config`
- Have `Config` extend `MinimalConfig`

That way, we'd have the same source of truth for the minimal config whether or not `with_minimal_config` is `true`, and we'd still be separating the contexts.

> Parallelly, I am working on a different way of handling configuration.

Maybe it would be best for us to discuss it first, otherwise you could be spending time on something that might not be agreed upon. Plus, I did some changes to `Config` in #109.

rafaelgomesxyz (changes requested):

lib/core/minimal_config.rb
ahangarha: Yes. MinimalConfig is a kind of super class for config. Perhaps the name should get changed to `Config` and `ExtendedConfig`.

I have another idea for the configuration class. I will share it with you.

justin808 (commented):
Could we solve this by loading the app config lazily and eliminate all the other crap?

ahangarha:
> Could we solve this by loading the app config lazily

I have worked on something like this (as side work). I work on it more to see if that implementation can fix the issue.

lib/cpl.rb
```diff
@@ -181,7 +182,9 @@ def self.all_base_commands
 
           show_info_header(config) if with_info_header
 
-          command_class.new(config).call
+          the_command = command_class.new(config)
+          the_command.prepare_extended_config unless with_minimal_config
+          the_command.call
```
rafaelgomesxyz, line 187: I don't really like this implementation.

Why don't we lazy-load the config like this: we change `apps`, `app`, `org` and `current` to be methods, and then if a method is called for the first time, we load the data needed.

For example, for app:

```rb
def app
  return @app if @app_loaded

  ...
  @app_loaded = true
end
```

Then we don't need `WITH_MINIMAL_CONFIG`, as it will just work everywhere.

rafaelgomesxyz (changes requested):

lib/cpl.rb
ahangarha: And how we control calling `load_app_config` and `load_apps` in `initialize` method of `Config` class? Those are the issue.

ahangarha: Done

ahangarha:
@justin808 
If this PR can get merged, let's do it because #116 is dependent on this branch and it is ready to be merged (if approved).

We can update the #109 based on this branch.

PR #112, opened 2023-11-20
Rename 'setup' to 'setup_app_templates' in config file

### LINKED ISSUE #96: Rename `controlplane.yml` key `setup` to `setup_app_workloads` for clarity
From #91 
---

ahangarha:
Note https://github.com/shakacode/heroku-to-control-plane/issues/96#issuecomment-1819318986

> I have two proposals here:
> 
> 1. Drop the `setup-app` command entirely and replace it with `--all` in `apply-template`.
>    I think the value of this command is to ease the initial setup of an app with all of its templates. If the user has any specific use case, then the normal usage of `apply-template t1 t2 t3` is there.
> 2. We assume `setup_app_template` includes all templates unless specified in the config file. This means if the user has not made an entry in the config file, still `setup-app` can be used, and it applies all the templates. In this case, the user can white-list templates in the config file under `setup_app_template`.
> 
> I think the first approach is better because it is much simpler.

---
Closes #96

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ _Please add entries here for your pull requests that are not yet released._
 ### Changed
 
 - `--org` option now takes precedence over `CPLN_ORG` env var, which takes precedence over `cpln_org` from `controlplane.yml`. [PR 88](https://github.com/shakacode/heroku-to-control-plane/pull/88) by [Rafael Gomes](https://github.com/rafaelgomesxyz).
+- Deprecated `setup` entry in the ControlePlane config file and introduced `setup_app_templates` in its place. [PR 112](https://github.com/shakacode/heroku-to-control-plane/pull/112) by [Mostafa Ahangarhga](https://github.com/ahangarha).
 
 ## [1.1.2] - 2023-10-17
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ _Please add entries here for your pull requests that are not yet released._
 ### Changed
 
 - `--org` option now takes precedence over `CPLN_ORG` env var, which takes precedence over `cpln_org` from `controlplane.yml`. [PR 88](https://github.com/shakacode/heroku-to-control-plane/pull/88) by [Rafael Gomes](https://github.com/rafaelgomesxyz).
-- Deprecated `setup` entry in the Control Plane config file and introduced `setup_app_templates` in its place. [PR 112](https://github.com/shakacode/heroku-to-control-plane/pull/112) by [Mostafa Ahangarhga](https://github.com/ahangarha).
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -170,7 +170,7 @@ aliases:
 
     # Allows running the command `cpl setup-app`
     # instead of `cpl apply-template gvc redis postgres memcached rails sidekiq`.
-    setup:
+    setup_app_templates:
```
rafaelgomesxyz, line 173: There are more places where we need to update this, such as in `migrating.md` and in the command description. Please search for occurrences of `setup` and replace with `setup_app_templates` wherever appropriate.

rafaelgomesxyz (changes requested):

justin808 (changes requested):
Need a deprecation message and CHANGELOG entry.

lib/core/config.rb
```diff
@@ -133,6 +133,7 @@ def new_option_keys
       org: :cpln_org,
       location: :default_location,
       prefix: :match_if_app_name_starts_with,
+      setup: :setup_app_templates,
```
ahangarha, line 136: @justin808 
We get the intended deprecation message:
```
DEPRECATED: Option 'setup' is deprecated, please use 'setup_app_templates' instead (in 'controlplane.yml').
```

README.md
ahangarha: Done

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ _Please add entries here for your pull requests that are not yet released._
 ### Changed
 
 - `--org` option now takes precedence over `CPLN_ORG` env var, which takes precedence over `cpln_org` from `controlplane.yml`. [PR 88](https://github.com/shakacode/heroku-to-control-plane/pull/88) by [Rafael Gomes](https://github.com/rafaelgomesxyz).
+- Deprecated `setup` entry in the ControlePlane config file and introduced `setup_app_templates` in its place. [PR 112](https://github.com/shakacode/heroku-to-control-plane/pull/112) by [Mostafa Ahangarhga](https://github.com/ahangarha).
```
rafaelgomesxyz, line 30: Typo:

```suggestion
- Deprecated `setup` entry in the Control Plane config file and introduced `setup_app_templates` in its place. [PR 112](https://github.com/shakacode/heroku-to-control-plane/pull/112) by [Mostafa Ahangarhga](https://github.com/ahangarha).
```

Also, I feel like "Renamed `setup` config to `setup_app_templates`." would be simpler and gets the point across.

rafaelgomesxyz (approved):
Just a typo in the changelog, the rest looks good to me.

CHANGELOG.md
ahangarha: I think it would be good to mention deprecation somewhere in the changelog.

PR #114, opened 2023-11-20
Show info header only on first command

### LINKED ISSUE #111: Show info header only once
Some commands (like `setup-app`) runs other commands. In such cases, we show the info header (for `org` and `app` name) twice.

Example:
```
➜ cpl setup-app
ORG: mostafa-demo
APP: tutorial-app

ORG: mostafa-demo
APP: tutorial-app

Applying template 'rails'...
```

ahangarha:
---
Closes #111

### DIFF
--- a/lib/cpl.rb
+++ b/lib/cpl.rb
@@ -44,7 +44,7 @@ module Cpl
     package_name "cpl"
     default_task :no_command
 
-    @@already_showed_info_header = false # rubocop:disable Style/ClassVars
+    @already_showed_info_header = false
 
     def self.start(*args)
       check_cpln_version
@@ -181,7 +181,7 @@ module Cpl
         begin
           config = Config.new(args, options)
 
-          show_info_header(config) if with_info_header
+          Cpl::Cli.show_info_header(config) if with_info_header
 
           command_class.new(config).call
         rescue RuntimeError => e
@@ -192,10 +192,8 @@ module Cpl
       ::Shell.abort("Unable to load command: #{e.message}")
     end
 
-    private
-
-    def show_info_header(config)
-      return if @@already_showed_info_header
+    def self.show_info_header(config)
+      return if @already_showed_info_header
 
       rows = {}
       rows["ORG"] = config.org || "NOT PROVIDED!"
@@ -205,7 +203,7 @@ module Cpl
         puts "#{key}: #{value}"
       end
 
-      @@already_showed_info_header = true # rubocop:disable Style/ClassVars
+      @already_showed_info_header = true
 
       # Add a newline after the info header
       puts
--- a/lib/cpl.rb
+++ b/lib/cpl.rb
@@ -44,7 +44,7 @@ module Cpl
     package_name "cpl"
     default_task :no_command
 
-    @already_showed_info_header = false
+    @showed_info_header = false
 
     def self.start(*args)
       check_cpln_version
@@ -193,7 +193,7 @@ module Cpl
     end
 
     def self.show_info_header(config)
-      return if @already_showed_info_header
+      return if @showed_info_header
 
       rows = {}
       rows["ORG"] = config.org || "NOT PROVIDED!"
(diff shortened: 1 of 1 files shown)


lib/cpl.rb
```diff
@@ -204,6 +206,10 @@ def show_info_header(config)
       # Add a newline after the info header
       puts
     end
+
+    def first_run?
+      ObjectSpace.each_object(self.class).count == 1
```
rafaelgomesxyz, line 211: Is this the best way to do this? In other places, we use instance/class variables:

omHjTIfJsh

rafaelgomesxyz (changes requested):

lib/cpl.rb
ahangarha: I am not happy about it either but this was the only way I could implement it.

I tried this at first. But when we run a command (run CLI start method) in another command, we get into a new instance, and we don't have access to the earlier instance variables.

justin808: This raises a WTF...

Where in the world did you get the ObjectSpace idea?

Counting object instances is a total hack.

You could probably set a **static** counter on or boolean on the CPL class. In other words, instance variable on the class, not the instance.

Use a class instance variable (aka static) on the cpl class.

https://chat.openai.com/share/e62c8db1-0622-4617-b8a6-5c3a4d82f387

justin808 (changes requested):

ahangarha:
In [b38664f](https://github.com/shakacode/heroku-to-control-plane/pull/114/commits/b38664f63439a65c3129c2bc9aa2ab2901e85e2f), I used class variable because I couldn't implement the solution using a class instance variable.

lib/cpl.rb
```diff
@@ -44,7 +44,7 @@ class Cli < Thor # rubocop:disable Metrics/ClassLength
     package_name "cpl"
     default_task :no_command
 
-    @@already_showed_info_header = false # rubocop:disable Style/ClassVars
+    @already_showed_info_header = false
```
rafaelgomesxyz, line 47: "already" feels redundant. `showed_info_header` already gets the point across.

lib/cpl.rb
```diff
@@ -181,7 +181,7 @@ def self.all_base_commands
         begin
           config = Config.new(args, options)
 
-          show_info_header(config) if with_info_header
+          Cpl::Cli.show_info_header(config) if with_info_header
```
rafaelgomesxyz, line 184: Can we not just call `show_info_header` directly without `Cpl::Cli`, since we're inside of the class?

rafaelgomesxyz (changes requested):

lib/cpl.rb
ahangarha: Actually, in the current implementation (where I used `show_info_header` as a static method), we don't need `Cpl::Cli`.

Thanks for reminding.

The reason I didn't use a class instance variable was that I couldn't find a way to keep `show_info_header` private and still be able to call it without an explicit `Cpl::Cli` prefix. I still don't think this method should be a static method.

lib/cpl.rb
```diff
@@ -44,6 +44,8 @@ class Cli < Thor # rubocop:disable Metrics/ClassLength
     package_name "cpl"
     default_task :no_command
 
+    @showed_info_header = false
```
rafaelgomesxyz, line 47: We don't need to initialize it.

lib/cpl.rb
rafaelgomesxyz: You're right.

Maybe it would be even better to move `show_info_header` to a helper then, because it would be easier to call it. But the current way works too, so up to you. I'll approve the PR.

rafaelgomesxyz (approved):
Looks good to me, just one thing I missed about initializing the variable, but no need for re-review.

lib/cpl.rb
ahangarha: Since we use it once, let's keep it here. Let's move it there when we need to. Agree?

justin808:
LGTM

justin808:
@ahangarha clear on why this is better?

ahangarha:
> clear on why this is better?

Sure!

PR #136, opened 2024-01-19
Fix issue that doesn't allow using upstream with `match_if_app_name_starts_with` set to `true` in `copy-image-from-upstream` command

rafaelgomesxyz:
**NOTE:** On top of #135 

The `copy-image-from-upstream` command does not currently allow using an app that has `match_if_app_name_starts_with` set to `true` as the upstream.

This PR fixes that.

### DIFF
--- a/lib/command/copy_image_from_upstream.rb
+++ b/lib/command/copy_image_from_upstream.rb
@@ -38,6 +38,7 @@ module Command
       pull_image_from_upstream
       push_image_to_app
     ensure
+      cp.profile_switch("default")
       delete_upstream_profile
     end
 
--- a/lib/command/copy_image_from_upstream.rb
+++ b/lib/command/copy_image_from_upstream.rb
@@ -29,7 +29,7 @@ module Command
       ensure_docker_running!
 
       @upstream = config[:upstream]
-      @upstream_org = config.apps[@upstream.to_sym][:cpln_org] || ENV.fetch("CPLN_ORG_UPSTREAM", nil)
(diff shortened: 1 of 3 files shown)


lib/core/config.rb
```diff
@@ -102,6 +102,13 @@ def current
     end
   end
 
+  def find_app_config(app_name1)
```
rafaelgomesxyz, line 105: Just made it public to be able to use it.

lib/command/copy_image_from_upstream.rb
```diff
@@ -29,7 +29,7 @@ def call # rubocop:disable Metrics/MethodLength
       ensure_docker_running!
 
       @upstream = config[:upstream]
-      @upstream_org = config.apps[@upstream.to_sym][:cpln_org] || ENV.fetch("CPLN_ORG_UPSTREAM", nil)
+      @upstream_org = config.find_app_config(@upstream)&.dig(:cpln_org) || ENV.fetch("CPLN_ORG_UPSTREAM", nil)
```
rafaelgomesxyz, line 32: Let's say we have a  `dummy-test` app in `controlplane.yml` that has `match_if_app_name_starts_with` set to `true`.

The old method wouldn't work if called with, e.g., `dummy-test-123`, because `config.apps` only contains the apps exactly as they are in `controlplane.yml`, and we have `dummy-test` in `controlplane.yml`, not `dummy-test-123`.

The new method works by iterating through `config.apps` looking for an app that matches `dummy-test-123`, so it will find `dummy-test` and match it.

justin808: I don't understand why we'd use a starts with for the upstream. That should be a validation error. Agree, @dzirtusss

PR #137, opened 2024-01-19
Change precedence for `CPLN_ORG_UPSTREAM` env var in `copy-image-from-upstream` command

rafaelgomesxyz:
**NOTE:** On top of #136 

Environment variables should take precedence over the `controlplane.yml` file. This is currently not true for the `CPLN_ORG_UPSTREAM` env var in the `copy-image-from-upstream` command.

This PR changes that.


justin808:
@rafaelgomesxyz double-check my merge commit.

PR #139, opened 2024-01-19
Updates to `info` command

rafaelgomesxyz:
**NOTE:** On top of #138 

Some updates to the `info` command:

- It now works properly for apps with `match_if_app_name_starts_with` set to `true`. Before, when calling it with, e.g., `dummy-test-123`, it would ignore the specific app and show the info for `dummy-test`. Now it properly shows the info for `dummy-test-123`.
- It now lists the workloads in the same order as `controlplane.yml` (just a visual improvement).

### DIFF
--- a/lib/command/info.rb
+++ b/lib/command/info.rb
@@ -32,7 +32,7 @@ module Command
       @missing_apps_workloads = {}
       @missing_apps_starting_with = {}
 
-      if config.app && !config.current[:match_if_app_name_starts_with]
+      if config.app && !config.should_app_start_with?(config.app)
         single_app_info
       else
         multiple_apps_info
@@ -41,20 +41,8 @@ module Command
 
     private
 
-    def app_matches?(app, app_name, app_options)
-      app == app_name.to_s || (app_options[:match_if_app_name_starts_with] && app.start_with?(app_name.to_s))
-    end
-
-    def find_app_options(app)
-      @app_options ||= {}
-      @app_options[app] ||= config.apps.find do |app_name, app_options|
-                              app_matches?(app, app_name, app_options)
-                            end&.last
-    end
-
     def find_workloads(app)
-      app_options = find_app_options(app)
-      return [] if app_options.nil?
+      app_options = config.find_app_config(app)
 
       (app_options[:app_workloads] + app_options[:additional_workloads] + [app_options[:one_off_workload]]).uniq
     end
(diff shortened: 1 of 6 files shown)


lib/command/info.rb
```diff
@@ -41,20 +41,8 @@ def call
...
-    def find_app_options(app)
-      @app_options ||= {}
-      @app_options[app] ||= config.apps.find do |app_name, app_options|
-                              app_matches?(app, app_name, app_options)
-                            end&.last
-    end
-
     def find_workloads(app)
-      app_options = find_app_options(app)
-      return [] if app_options.nil?
```
rafaelgomesxyz, line 57: This line would never be executed, because this method is only called for apps that we're 100% sure are in `controlplane.yml`.

lib/command/info.rb
```diff
@@ -75,21 +63,19 @@ def fetch_app_workloads(org) # rubocop:disable Metrics/MethodLength
...
 
-    def orgs # rubocop:disable Metrics/MethodLength
+    def orgs
       result = []
 
       if config.org
         result.push(config.org)
       else
-        config.apps.each do |app_name, app_options|
-          next if config.app && !app_matches?(config.app, app_name, app_options)
```
rafaelgomesxyz, line 91: This line would never be executed, because providing an `app` without an `org` is not possible (an `app` must have an `org`).

lib/command/info.rb
```diff
@@ -184,8 +170,7 @@ def print_missing_apps_starting_with
            "(replace 'whatever' with whatever suffix you want):"
 
       @missing_apps_starting_with.each do |app, _workloads|
-        app_with_suffix = "#{app}#{app.end_with?('-') ? '' : '-'}whatever"
-        puts "  - `cpl setup-app -a #{app_with_suffix}`"
+        puts "  - `cpl setup-app -a #{app}-whatever`"
```
rafaelgomesxyz, line 173: We don't need to do this check, we can just add the suffix.

justin808: What is the "whatever" I think we can give a better message.

justin808:
```
puts " - `cpl setup-app -a #{app}-pr-1234`"
```

justin808 (changes requested):

lib/command/info.rb
justin808: OK -- I see you defined whatever above.

PR #141, opened 2024-02-28
feat: detect bad outgoing links

samueldemoura:
## Summary

Implements automatic checks to ensure we're linking to the correct URLs.

Checking is done via `git grep` with a small bash script, and there's a corresponding GitHub Action workflow to run it.

### DIFF
--- a/.github/workflows/check_cpln_links.yml
+++ b/.github/workflows/check_cpln_links.yml
@@ -12,7 +12,7 @@ jobs:
     name: Check Links
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Validate outgoing links to Control Plane
         run: |
           TERM=xterm-color ./script/check_cpln_links
--- a/.github/workflows/check_cpln_links.yml
+++ b/.github/workflows/check_cpln_links.yml
@@ -1,0 +1,19 @@
+name: Check Links
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  check_cpln_links:
+    runs-on: ubuntu-latest
+    name: Check Links
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Validate outgoing links to Control Plane
+        run: |
+          ./script/check_cpln_links
+        shell: bash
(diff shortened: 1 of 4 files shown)


.github/workflows/check_cpln_links.yml
```diff
@@ -0,0 +1,19 @@
...
+      - main
+  pull_request:
+
+jobs:
+  check_cpln_links:
+    runs-on: ubuntu-latest
+    name: Check Links
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
```
rafaelgomesxyz, line 15: Let's use v3: `actions/checkout@v3`

There's actually an even newer version (v4), but let's at least keep it consistent with the other actions.

.github/workflows/check_cpln_links.yml
```diff
@@ -0,0 +1,19 @@
...
+      - name: Validate outgoing links to Control Plane
+        run: |
+          TERM=xterm-color ./script/check_cpln_links
+        shell: bash
```
rafaelgomesxyz, line 19: Is this required?

It does say [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell) that if `bash` is not specified explicitly and it's not found in the path, it treats it as `sh` instead, but does this mean that we don't have `bash` in the path here?

I guess just remove if not needed.

rafaelgomesxyz (approved):
LGTM. Just a couple comments.

.github/workflows/check_cpln_links.yml
samueldemoura: Oops, good catch. Done in 7e02edf.

.github/workflows/check_cpln_links.yml
samueldemoura: > Is this required?

Dunno. The example for running a Bash command includes it, and I personally use `fish` so after running into enough bugs due to shebang-less scripts, I've gotten into the habit of being explicit about which shell I'm expecting any time I'm running something.

I prefer keeping the explicit shell selection, but if you feel strongly about it, let me know and I can drop it.

rafaelgomesxyz (approved):

PR #147, opened 2024-03-24
Run release script using cpl run command

justin808:
Running in the context of the CI is incorrect. Needs to use the deployed image latest.

Plus, doc improvements.

Match the location link with current yaml.


rafaelgomesxyz:
Some of these changes will be integrated into https://github.com/shakacode/heroku-to-control-plane/pull/151. Closing for now.

PR #150, opened 2024-04-02
cpl doctor command, including validation for duplicate app matching when prefix

rafaelgomesxyz:
Add controlplane.yml verification, first check is invalid prefixes that collide with other apps or prefixes

There's currently a bug where if there's an app defined after another app (both with `match_if_app_name_starts_with` set to `true`), and the name of the second app is a prefix in the name of the first app, and we try to run commands for apps with a name matching the first app, the second app will be used instead.

For example, let's say that we have this in `.controlplane/controlplane.yml`:

```yml
app-test-other:
  <<: *common
  match_if_app_name_starts_with: true

app-test:
  <<: *common
  match_if_app_name_starts_with: true
```

If we call a command like `cpl deploy-image -a app-test-other-123`, it will use the config from `app-test` instead of the one from `app-test-other`.

This PR fixes that by immediately returning an exit code of 1 from any cpl command.

Maybe add a `cpl doctor` command, like `brew doctor`, giving helpful tips and verifying configuration.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ Changes since the last non-beta release.
 
 _Please add entries here for your pull requests that are not yet released._
 
+### Fixed
+
+- Fixed app matching when name starts with (the wrong config was being used in some cases - see PR for more details). [PR 150](https://github.com/shakacode/heroku-to-control-plane/pull/150) by [Rafael Gomes](https://github.com/rafaelgomesxyz).
+
 ## [1.4.0] - 2024-03-20
 
 ### Added
--- a/lib/core/config.rb
+++ b/lib/core/config.rb
@@ -116,11 +116,10 @@ class Config # rubocop:disable Metrics/ClassLength
   def find_app_config(app_name1)
     @app_configs ||= {}
 
-    @app_configs[app_name1] ||= apps.filter_map do |app_name2, app_config|
+    @app_configs[app_name1] ||= apps.find do |app_name2, app_config|
                                   next unless app_matches?(app_name1, app_name2, app_config)
 
                                   app_config[:name] = app_name2
-                                  app_config
                                 end&.last
   end
 


justin808:
Change to simply show error in this case where **any prefix** is included in any other app name (including prefixes).

Thus, we need some controlplane.yml verifier, and here's our first verification check: 

```ruby
def app_name_contained_in_other_app_name(app_name_list, app_prefix_list)
  sorted_list = app_name_list.sort
  matches = []
  app_prefix_list.each do |prefix|
    index = sorted_list.bsearch_index { |app_name| prefix <= app_name }
    next unless index
    next_app_name = sorted_list[index]
    if next_app_name.start_with?(prefix) && next_app_name != prefix
      matches << "'#{prefix}' is a prefix of '#{next_app_name}'"
    end
  end
  matches.empty? ? nil : matches.join(", ")
end
```

Tests:

```ruby
require 'rspec'

describe '#app_name_contained_in_other_app_name' do
  it 'returns an informative message when a match is found' do
    result = app_name_contained_in_other_app_name(['app-test-other', 'app-test-again'], ['app-test'])
    expect(result).to eq("'app-test' is a prefix of 'app-test-other', 'app-test' is a prefix of 'app-test-again'")
  end

  it 'returns nil when no match is found' do
    result = app_name_contained_in_other_app_name(['app-review'], ['app-storybook'])
    expect(result).to be_nil
  end

  it 'returns nil when the prefix is the same as the app name' do
    result = app_name_contained_in_other_app_name(['short-name', 'short-name-etc'], ['short-name'])
    expect(result).to be_nil
  end

  it 'returns an informative message when the prefix is a prefix of the app name' do
    result = app_name_contained_in_other_app_name(['short-name-etc', 'short-name-etc-long'], ['short-name-etc'])
    expect(result).to eq("'short-name-etc' is a prefix of 'short-name-etc-long'")
  end

  it 'returns all matches when multiple prefixes match' do
    result = app_name_contained_in_other_app_name(['app-test-other', 'app-test-again', 'short-name-etc-long'], ['app-test', 'short-name-etc'])
    expect(result).to eq("'app-test' is a prefix of 'app-test-other', 'app-test' is a prefix of 'app-test-again', 'short-name-etc' is a prefix of 'short-name-etc-long'")
  end
end
```

rafaelgomesxyz:
Will open a separate PR for `cpl doctor` to keep things clean.

PR #151, opened 2024-04-03
Updates to `run` command

rafaelgomesxyz:
- Changes the `run` command to use a Cron workload, having both non-interactive and interactive modes
- Adds options to override the workload container's `--cpu`, `--memory`, and `--entrypoint`
- Adds an option to delete a specific workload to the `delete` command
- Adds an option to view logs from a specific replica to the `logs` command
- Adds an option to stop a specific replica to the `ps:stop` command
- Deprecates the `run:detached` command in favor of `run` (this is fully backwards-compatible, as the `run` command now operates on a non-interactive and non-detached mode by default)
- Removes the `run:cleanup` command, as we now use a single reusable workload
- Changes the release script in the `deploy-image` command to run in the context of the `run` command

## Examples

- `cpl run -a app`: Runs `bash` by default, interactively.
- `cpl run -a app --interactive -- cmd`: Runs `cmd` interactively (if `cmd` is `bash` or `rails c` etc, no need to pass `--interactive` - it detects it automatically).
- `cpl run -a app --detached -- cmd`: Executes `cmd`, exits, and prints commands for seeing logs and stopping the job.
- `cpl run -a app -- cmd`: Executes `cmd` and starts outputting logs (Ctrl+C detaches and prints the same commands as above). **Returns non-zero exit code when `cmd` fails (has non-zero exit code)**, so this is used for remote execution. We do not return the exact exit code from failed `cmd`. That is likely in the logs and can be a future enhancement.
- `cpl logs -a app --workload <workload> --replica <replica>`: Outputs logs for a detached job.
- `cpl ps:stop -a app --workload <workload> --replica <replica>`: Stops a detached job.

## Summary by CodeRabbit

- **New Features**
  - Enhanced management of workloads and Docker images through refined command functionalities.
  - Added support for interactive and non-interactive replicas in command executions.
  - Streamlined database tasks in build processes and deployment strategies.
  - Improved user guidance and command descriptions in documentation and example configurations.

- **Bug Fixes**
  - Adjusted database migration commands to remove the `:detached` flag for more streamlined operations.

- **Documentation**
  - Updated usage instructions and command descriptions across README, migration guides, and command documentation.
  - Clarified and expanded command functionalities and parameters in user guides.

- **Refactor**
  - Refined internal command structures and option handling for better performance and user experience.
  - Updated method names and parameters for clarity and improved functionality in workload management.

- **Tests**
  - Enhanced testing frameworks to cover new functionalities and command behaviors.
  - Added new test contexts and scenarios for comprehensive coverage.

- **Chores**
  - Simplified command structures and removed deprecated commands for cleaner codebase and improved maintenance.


lib/command/run_logs.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
+
+module Command
+  class RunLogs < Base
```
rafaelgomesxyz, line 4: The code here is essentially the same that was in `run:detached` for fetching the logs, with minor modifications, like getting the exit code from the command.

lib/core/scripts.rb
```diff
@@ -1,34 +0,0 @@
...
-  end
-
-  def helpers_cleanup
-    <<~SHELL
-      unset CONTROLPLANE_RUNNER
-    SHELL
-  end
-
-  # NOTE: please escape all '/' as '//' (as it is ruby interpolation here as well)
-  def http_dummy_server_ruby
```
rafaelgomesxyz, line 26: This was the only method that was being used from this module, and I don't understand why we can't just use `tail -f /dev/null`, so I did that in `run` and dropped this.

docs/commands.md
```diff
@@ -105,11 +105,15 @@ cpl copy-image-from-upstream -a $APP_NAME --upstream-token $UPSTREAM_TOKEN --ima
```
rafaelgomesxyz, line 105: All modifications in this file come from the command files, so it might be best to always ignore this and just go straight to the source (this file is always automatically generated based on the source).

lib/command/run.rb
```diff
@@ -51,59 +69,89 @@ class Run < Base
       ```
     EX
 
-    attr_reader :location, :workload_to_clone, :workload_clone, :container
+    INTERACTIVE_COMMANDS = [
+      "bash",
+      "rails console",
+      "rails c",
+      "rails dbconsole",
+      "rails db"
```
rafaelgomesxyz, line 77: Any others?

lib/command/run.rb
```diff
@@ -51,59 +69,89 @@ class Run < Base
...
-      # Needed to avoid execution of ENTRYPOINT and CMD of Dockerfile
-      container_spec["command"] = "ruby"
-      container_spec["args"] = ["-e", Scripts.http_dummy_server_ruby]
+      # Set runner
+      container_spec["env"] ||= []
+      container_spec["env"] << { "name" => "CONTROLPLANE_RUNNER", "value" => runner_script }
+      if interactive
+        container_spec["command"] = "tail"
+        container_spec["args"] = ["-f", "/dev/null"]
+        @command = %(bash -c 'eval "$CONTROLPLANE_RUNNER"')
```
rafaelgomesxyz, line 130: Using `tail -f /dev/null` as the default command for interactive mode allows the cron workload to stay active indefinitely, so that we can connect to it later and run the script using `cpln workload exec`.

lib/command/run.rb
```diff
@@ -51,59 +69,89 @@ class Run < Base
...
+      else
+        container_spec["command"] = "bash"
+        container_spec["args"] = ["-c", 'eval "$CONTROLPLANE_RUNNER"']
```
rafaelgomesxyz, line 133: For non-interactive mode, we don't need to keep the cron workload active, so we just run the script directly.

lib/command/run.rb
```diff
@@ -51,59 +69,89 @@ class Run < Base
...
+      end
 
-      # Ensure one-off workload will be running
-      spec["defaultOptions"]["suspend"] = false
+      # Ensure cron workload won't run per schedule
+      spec["defaultOptions"]["suspend"] = true
```
rafaelgomesxyz, line 137: This ensures that the cron workload won't be triggered by the schedule and will stay disabled until we manually run it with `cpln workload cron start`.

lib/command/run.rb
```diff
@@ -136,10 +196,52 @@ def runner_script # rubocop:disable Metrics/MethodLength
       script
     end
 
-    def run_in_replica
+    def non_interactive_runner_script
```
rafaelgomesxyz, line 199: The code here is essentially the same that was in `run:detached` for the runner script.

justin808:
Kyle Cupp
  [5 hours ago](https://shakacode.slack.com/archives/C03RWL8EP5Y/p1714054030712109?thread_ts=1713914065.458379&cid=C03RWL8EP5Y)
A replica and a job in this context are the same thing. So yes, they both stop simultaneously.
Aside from the UI, the best way to stop a replica is using [this command](https://docs.controlplane.com/reference/cli#workload-replica-stop)
Nothing happens to a running job when the workload image changes.

docs.controlplane.comdocs.controlplane.com
[Control Plane - From many clouds, one™](https://docs.controlplane.com/reference/cli#workload-replica-stop)
Serve microservices in an unbreakable, low-latency manner.

lib/command/base.rb
```diff
@@ -294,6 +352,30 @@ def wait_for_replica(workload, location)
...
+        status = result&.dig("items", 0)&.dig("status")
+
+        case status
+        when :invalid, :removed, :failure
+          return 1
+        when :success
+          return 0
+        end
+
+        Kernel.sleep(1)
```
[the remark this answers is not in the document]
rafaelgomesxyz: By default cron jobs run indefinitely. Do we want to add a forced timeout or add an option to specify timeout?

If we add a timeout to the cron job, we don't need to add a timeout here, because this loop finishes once the job no longer exists.

dzirtusss (changes requested):
I guess overall LGTM except changes.

What I'm worry more is "real-life testing". I guess at least it needs:

1. have a thorough testing of typical bash/rails commands:
```
bash
bash -c ls # short
bash -c sleep 10000 # long
rails c # and do something
rails db:migrate:status # short
rails runner "sleep(10000)" # long
```
2. have a long-time testing of above with say ~20min to double check that such commands are not interrupted

3. run several long commands in parallel shells

4. run long commad, deploy new image, run second command (with new image). Actually I dunno what will happen in such a case. With previous logic, it worked and first command was running with old image, and new command was running with new image. Both simultaneously.
Why this might be needed? Say I start `run:detached` with some many hours job. Then some dev trigger deploy.
I mean if this will not work, it should be clearly stated in docs that this doesn't work.

So... I guess testing is key for this PR.

As well sure various new options (e.g. entrypoint) should be tested to work as expected.

.github/workflows/rspec.yml
```diff
@@ -5,36 +5,24 @@ on:
...
-        if: always()
-        with:
-          name: coverage-report-${{ github.run_id }}-${{ matrix.os }}-${{ matrix.ruby }}
-          path: coverage
+  rspec-fast:
+    name: RSpec (Fast)
+    uses: ./.github/workflows/rspec-shared.yml
+    with:
+      os-version: ubuntu-latest
+      ruby-version: "3.0"
```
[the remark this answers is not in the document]
rafaelgomesxyz: This is on top of the other branch, so it will be fixed with a rebase.

lib/command/delete.rb
```diff
@@ -5,28 +5,54 @@ class Delete < Base
...
+    DESCRIPTION = "Deletes the whole app (GVC with all workloads, all volumesets and all images) or a specific workload"
     LONG_DESCRIPTION = <<~DESC
-      - Deletes the whole app (GVC with all workloads, all volumesets and all images)
+      - Deletes the whole app (GVC with all workloads, all volumesets and all images) or a specific workload
       - Will ask for explicit user confirmation
     DESC
+    EXAMPLES = <<~EX
+      ```sh
+      # Deletes the whole app (GVC with all workloads, all volumesets and all images).
+      cpl delete -a $APP_NAME
```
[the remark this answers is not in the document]
rafaelgomesxyz: Looks like Thor defaults to the same name as the option:

example

So it will look for a workload named "workload".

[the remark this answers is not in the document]
justin808: Please validate the name exists and is not `workload`

lib/command/run.rb
```diff
@@ -48,80 +63,178 @@ class Run < Base
...
 
       # Ensure no scaling
-      spec["defaultOptions"]["autoscaling"]["minScale"] = 1
-      spec["defaultOptions"]["autoscaling"]["maxScale"] = 1
+      spec["defaultOptions"]["autoscaling"] = {}
       spec["defaultOptions"]["capacityAI"] = false
 
+      # Set cron job props
+      spec["type"] = "cron"
+      spec["job"] = { "schedule" => "* * * * *", "restartPolicy" => "Never" }
```
justin808, line 163:
```suggestion
      spec["job"] = { "schedule" => "0 0 1 1 1 ", "restartPolicy" => "Never" }
```

Maybe use
> You can set the Schedule for 0 0 1 1 1 , which means the next Job will run on January 1 2029.

lib/command/run.rb
```diff
@@ -48,80 +63,178 @@ class Run < Base
...
       image = config.options[:image]
       image = latest_image if image == "latest"
       container_spec["image"] = "/org/#{config.org}/image/#{image}" if image
 
-      # Set runner
-      container_spec["env"] ||= []
-      container_spec["env"] << { "name" => "CONTROLPLANE_RUNNER", "value" => runner_script }
+      # Container overrides
+      container_spec["cpu"] = config.options[:cpu] if config.options[:cpu]
+      container_spec["memory"] = config.options[:memory] if config.options[:memory]
```
justin808, line 182: we should have some basic validations of CPU and memory

lib/command/run.rb
```diff
@@ -48,80 +63,178 @@ class Run < Base
...
-      end
-
-      # Create workload clone
+      # Update workload clone
       cp.apply_hash("kind" => "workload", "name" => workload_clone, "spec" => spec)
     end
 
-    def runner_script # rubocop:disable Metrics/MethodLength
-      script = Scripts.helpers_cleanup
+    def build_data
```
justin808, line 188:
```suggestion
    def build_job_start_yaml
```

lib/command/run.rb
```diff
@@ -48,80 +63,178 @@ class Run < Base
...
+      data = <<~DATA
```
justin808, line 189:
```suggestion
      job_start_yaml = <<~JOB_START_YAML
```

lib/command/run.rb
```diff
@@ -48,80 +63,178 @@ class Run < Base
...
-      ensure_workload_deleted(workload_clone)
+      if interactive
+        run_interactive
+      else
+        run_non_interactive
+      end
+    ensure
+      if interactive
+        progress.puts
+        stop_replica(workload_clone, replica, location)
```
justin808, line 127: It's important that we have a timeout for interactive commands in case the program quits.

Can we have the main job program be something other than `tail -f` so it can check if anybody is connected to it and start checking that after a few minutes, and if no other process is connected, then issue command to terminate itself.

justin808: ask for help from control plane channel on how to do this.

lib/command/run.rb
```diff
@@ -48,80 +63,178 @@ class Run < Base
...
-          CPLN_TOKEN=$CONTROLPLANE_TOKEN
-          unset CONTROLPLANE_TOKEN
-        SHELL
+      if interactive
+        data += <<~COMMAND
+          #{"command: #{command}" if command}
+          args:
+            #{'- "bash"' unless command == "bash"}
+            - "-c"
+            - 'eval "tail -f /dev/null"'
```
justin808, line 218: See prior comment. We need some sort of job that monitors the interactive connection.

lib/command/run.rb
```diff
@@ -48,80 +63,178 @@ class Run < Base
...
+          COMMAND
+        @command = %(bash -c 'eval "#{args_join(config.args)}"')
+      else
+        data += <<~COMMAND
+          #{"command: #{command}" if command}
+          args:
+            #{'- "bash"' unless command == "bash"}
+            - "-c"
+            - 'eval "#{args_join(config.args)}"'
```
justin808, line 227: This `eval` needs explaining.

justin808: ask ai for help

lib/core/controlplane.rb
```diff
@@ -225,13 +221,29 @@ def workload_connect(workload, location:, container: nil, shell: nil)
...
-  def workload_exec(workload, location:, container: nil, command: nil)
-    cmd = "cpln workload exec #{workload} #{gvc_org} --location #{location}"
+  def workload_exec(workload, replica, location:, container: nil, command: nil)
+    cmd = "cpln workload exec #{workload} #{gvc_org} --replica #{replica} --location #{location}"
     cmd += " --container #{container}" if container
     cmd += " -- #{command}"
     perform!(cmd, output_mode: :all)
   end
 
+  def workload_cron_start(workload, data, location:)
```
justin808, line 231:
```suggestion
  def workload_cron_start(workload, job_start_yaml, location:)
```

lib/core/controlplane.rb
```diff
@@ -225,13 +221,29 @@ def workload_connect(workload, location:, container: nil, shell: nil)
...
+    Tempfile.create do |f|
+      f.write(data)
```
justin808, line 233:
```suggestion
      f.write(job_start_yaml)
```

lib/core/controlplane.rb
```diff
@@ -286,8 +298,11 @@ def set_domain_workload(data, workload)
 
   # logs
 
-  def logs(workload:)
-    cmd = "cpln logs '{workload=\"#{workload}\"}' --org #{org} -t -o raw --limit 200"
+  def logs(workload:, replica: nil)
+    query_parts = ["workload=\"#{workload}\""]
+    query_parts.push("replica=\"#{replica}\"") if replica
+
+    cmd = "cpln logs '{#{query_parts.join(",")}}' --org #{org} -t -o raw --limit 200"
```
justin808, line 305: 200 is a magic number

make the limit configurable

justin808 (changes requested):

rafaelgomesxyz:
Pending:

- [ ] Remove start/stop hooks
- [ ] Add timeout option
- [ ] Add options validation
- [ ] Add monitoring script for interactive mode
- [ ] Tests

spec/command/logs_spec.rb
```diff
@@ -4,36 +4,125 @@
...
+
+      expect(message_result).to include("Fetching logs for workload 'postgres'")
+      expect(logs_result).to match(logs_regex)
+    end
+  end
+
+  context "when replica is provided" do
+    let!(:replica) do
+      run_cpl_command!("ps:stop", "-a", app, "--wait")
+      run_cpl_command!("ps:start", "-a", app, "--wait")
```
rafaelgomesxyz, line 50: Need to stop and start to make sure there aren't any leftover replicas when running `ps` below.

spec/command/ps_stop_spec.rb
```diff
@@ -9,29 +9,58 @@
...
+      expect(result[:stderr]).not_to include("postgres")
+    end
   end
 
-  it "stops all workloads and waits for them to not be ready", :slow do
-    result = run_cpl_command("ps:stop", "-a", app, "--wait")
+  context "when replica is provided" do
+    let!(:replica) do
+      run_cpl_command!("ps:stop", "-a", app, "--wait")
+      run_cpl_command!("ps:start", "-a", app, "--wait")
```
rafaelgomesxyz, line 45: Need to stop and start to make sure there aren't any leftover replicas when running `ps` below.

lib/core/controlplane.rb
```diff
@@ -143,40 +143,38 @@ def query_workloads(workload, a_gvc = gvc, a_org = org, partial_workload_match:
     api.query_workloads(org: a_org, gvc: a_gvc, workload: workload, gvc_op_type: gvc_op, workload_op_type: workload_op)
   end
 
-  def workload_get_replicas(workload, location:)
-    cmd = "cpln workload get-replicas #{workload} #{gvc_org} --location #{location} -o yaml"
+  def fetch_workload_replicas(workload, location:)
+    cmd = "cpln workload replica get #{workload} #{gvc_org} --location #{location} -o yaml"
```
rafaelgomesxyz, line 147: `replica get` is same as `get-replicas` (the latter was deprecated for the former).

lib/core/controlplane.rb
```diff
@@ -143,40 +143,38 @@ def query_workloads(workload, a_gvc = gvc, a_org = org, partial_workload_match:
...
-      ready = container.dig("resources", "replicas") == container.dig("resources", "replicasReady")
-      expected_status == true ? ready : !ready
+      container.dig("resources", "replicas") == container.dig("resources", "replicasReady")
     end
   end
 
-  def workload_deployments_ready?(workload, expected_status:)
+  def workload_deployments_ready?(workload, location:, expected_status:)
+    deployed_replicas = fetch_workload_replicas(workload, location: location)["items"].length
+    return deployed_replicas.zero? if expected_status == false
```
rafaelgomesxyz, line 170: With the previous check, sometimes it would return `true` even though there were still replicas deployed when waiting for the workload to not be ready. Now we check explicitly that the number of deployed replicas is zero, which is more accurate.

lib/command/run.rb
```diff
@@ -48,79 +74,241 @@ class Run < Base
...
+          exit(ExitCode::SUCCESS)
+        end
+
+        Cpl::Cli.start(["logs", *app_workload_replica_args])
+      end
+      Process.detach(logs_pid)
+
+      # We need to wait a bit for the logs to appear,
+      # otherwise it may exit without showing them
+      Kernel.sleep(30)
```
rafaelgomesxyz, line 237: Perhaps there's a better way to ensure that all logs have been shown instead of waiting an arbitrary amount of seconds?

lib/command/run.rb
```diff
@@ -48,79 +74,241 @@ class Run < Base
...
+      end
+
+      job_start_hash["command"] = entrypoint if entrypoint
+      job_start_hash["args"] ||= []
+      job_start_hash["args"].push("bash") unless entrypoint == "bash"
+      job_start_hash["args"].push("-c")
+      if interactive
+        job_start_hash["env"] ||= []
+        job_start_hash["env"].push({ "name" => "CPL_RUNNER_SCRIPT", "value" => interactive_runner_script })
+        job_start_hash["env"].push({ "name" => "CPL_MONITORING_SCRIPT", "value" => interactive_monitoring_script })
```
rafaelgomesxyz, line 273: These need to be set as env vars to avoid quote escaping issues, but since they're set on a job-basis, this won't interfere with other jobs.

lib/command/run.rb
```diff
@@ -48,79 +74,241 @@ class Run < Base
...
+        job_start_hash["args"].push('eval "$CPL_MONITORING_SCRIPT"')
+        @command = %(bash -c 'eval "$CPL_RUNNER_SCRIPT"')
+      else
+        job_start_hash["args"].push(%(eval "#{args_join(config.args)}"))
+      end
+
+      job_start_hash.to_yaml
+    end
+
+    def interactive_monitoring_script
```
rafaelgomesxyz, line 284: Could be a constant maybe?

lib/command/run.rb
```diff
@@ -48,86 +74,248 @@ class Run < Base
...
+    def interactive_runner_script # rubocop:disable Metrics/MethodLength
+      script = ""
 
       # NOTE: fixes terminal size to match local terminal
       if config.current[:fix_terminal_size] || config.options[:terminal_size]
         if config.options[:terminal_size]
           rows, cols = config.options[:terminal_size].split(",")
         else
-          rows, cols = Shell.cmd("stty", "size")[:output].split(/\s+/)
+          rows, cols = `stty size`.split(/\s+/)
```
rafaelgomesxyz, line 318: `stty size` has to be run in a terminal environment, so we can't use `Shell.cmd` because it uses `Open3.capture`.

lib/command/run.rb
rafaelgomesxyz: `eval` makes it easier to run multiline scripts and circumvent some path issues. Since this works, how about we leave it like this for now and investigate potentially removing `eval` in a follow up PR?

lib/cpl.rb
```diff
@@ -210,6 +215,10 @@ def self.all_base_commands
       ::Shell.abort("Unable to load command: #{e.message}")
     end
 
+    disable_required_check!(*@commands_with_required_options)
+    check_unknown_options!(except: @commands_with_extra_options)
+    stop_on_unknown_option!
```
rafaelgomesxyz, line 220: This ensures that an error is raised if unknown options are passed to any command, with the exception of commands that accept extra options and pass those forward.

lib/command/run.rb
dzirtusss: overall, IMO, it is worth extracting all such bash scripts to a dedicated module. just for clarity

lib/command/run.rb
dzirtusss: worth a `NOTE:`

docs/migrating.md
```diff
@@ -93,7 +93,7 @@ cpl setup-app -a my-app-staging
 cpl build-image -a my-app-staging --commit 456
 
 # Prepare database.
-cpl run:detached -a my-app-staging --image latest -- rails db:prepare
+cpl run -a my-app-staging --image latest -- rails db:prepare
```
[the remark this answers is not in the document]
rafaelgomesxyz: As I mentioned on Slack, the entrypoint should check that the Postgres service is running, not that the database exists, so I think it's fine to leave it like this everywhere.

dzirtusss (approved):
Overall LGTM

docs/migrating.md
```diff
@@ -215,9 +215,9 @@ fi
 
 # The `NEW_APP` env var that we exported above can be used to either reset or migrate the database before deploying.
 if [ -n "${NEW_APP}" ]; then
-  cpl run:detached 'LOG_LEVEL=warn rails db:reset' -a ${APP_NAME} --image latest
+  cpl run -a ${APP_NAME} --image latest -- rails db:reset
```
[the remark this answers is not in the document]
justin808: incorrect suggestion as `|| echo` does not error exist code.


🤯 🤯 🤯

spec/command/ps_stop_spec.rb
```diff
@@ -9,29 +9,58 @@
...
+
+    it "stops all workloads and waits for them to not be ready", :slow do
+      result = run_cpl_command("ps:stop", "-a", app, "--wait")
+
+      expect(result[:status]).to eq(0)
+      expect(result[:stderr]).to match(/Stopping workload 'rails'[.]+? done!/)
+      expect(result[:stderr]).to match(/Stopping workload 'postgres'[.]+? done!/)
+      expect(result[:stderr]).to match(/Waiting for workload 'rails' to not be ready[.]+? done!/)
+      expect(result[:stderr]).to match(/Waiting for workload 'postgres' to not be ready[.]+? done!/)
+    end
```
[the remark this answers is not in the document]
justin808: yes

justin808 (approved):

PR #152, opened 2024-04-04
Improve template parsing in `apply-template` command

rafaelgomesxyz:
This PR:

- Improves template parsing in the `apply-template` command when handling templates with multiple items in the same file
- Changes the `apply-template` command to raise an error if duplicate templates are found

### DIFF
--- a/lib/command/apply_template.rb
+++ b/lib/command/apply_template.rb
@@ -76,6 +76,10 @@ module Command
       end
     end
 
+    def template_duplicate?(template1, template2)
+      template1["kind"] == template2["kind"] && template1["name"] == template2["name"]
+    end
+
     def ensure_templates!
       missing_templates = config.args.reject { |name| File.exist?(template_filename(name)) }
       return if missing_templates.empty?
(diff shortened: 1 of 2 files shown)


lib/command/apply_template.rb
```diff
@@ -87,24 +65,83 @@ def call # rubocop:disable Metrics/CyclomaticComplexity, Metrics/MethodLength, M
...
       raise "Can't find templates above, please create them."
     end
 
+    def parse_templates # rubocop:disable Metrics/MethodLength
+      config.args.each_with_object([]) do |name, templates|
+        data = File.read(template_filename(name))
+        data = replace_variables(data)
+        templates_data = data.split(/^---\s*$/)
+        templates_data.each do |template_data|
+          template = YAML.safe_load(template_data)
```
rafaelgomesxyz, line 101: Before, we made assumptions about templates, like assuming that only a file named `gvc.yml` could be for an app, or that any other file would be for a workload.

Now, by parsing each item in each file separately, we can check its `kind` and be sure of what is an app and what is a workload.

And now we can use `YAML.safe_load`.

lib/command/apply_template.rb
```diff
@@ -40,40 +40,18 @@ class ApplyTemplate < Base # rubocop:disable Metrics/ClassLength
...
-
-      pending_templates = templates.select do |template|
-        if template == "gvc"
-          confirm_app(template)
-        else
-          confirm_workload(template)
-        end
-      end
-
-      progress.puts if @asked_for_confirmation
```
rafaelgomesxyz, line 60: Lines 50-60 moved to `confirm_templates` method.

lib/command/apply_template.rb
```diff
@@ -40,40 +40,18 @@ class ApplyTemplate < Base # rubocop:disable Metrics/ClassLength
...
-          items = apply_template(filename)
-          if items
-            items.each do |item|
-              report_success(item)
-            end
-          else
-            report_failure(template)
-          end
-
-          $CHILD_STATUS.success?
```
rafaelgomesxyz, line 75: Lines 65-75 moved to `apply_template` method.

rafaelgomesxyz:
Will be a part of `cpl doctor`.

PR #153, opened 2024-04-04
Raise error if image does not exist in `deploy-image` command

rafaelgomesxyz:
This PR changes the `deploy-image` command to to raise an error if the image does not exist.

### DIFF
--- a/lib/command/deploy_image.rb
+++ b/lib/command/deploy_image.rb
@@ -20,7 +20,7 @@ module Command
       deployed_endpoints = {}
 
       image = latest_image
-      raise "Image '#{image}' does not exist." if cp.fetch_image(image).nil?
+      raise "Image '#{image}' does not exist." if cp.fetch_image_details(image).nil?
 
       config[:app_workloads].each do |workload|
         workload_data = cp.fetch_workload!(workload)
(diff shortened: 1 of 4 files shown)


lib/command/deploy_image.rb
```diff
@@ -20,6 +20,7 @@ def call # rubocop:disable Metrics/MethodLength
       deployed_endpoints = {}
 
       image = latest_image
+      raise "Image '#{image}' does not exist." if cp.fetch_image(image).nil?
```
justin808, line 23: better error like this

```ruby
raise "Image '#{image}' does not exist in docker images on Control Plane https://console.cpln.io/console/org/hichee-production/repository/hichee-production. Use `cpl build-image` first." if cp.fetch_image(image).nil?
```

lib/core/controlplane.rb
```diff
@@ -58,6 +58,10 @@ def image_build(image, dockerfile:, docker_args: [], build_args: [], push: true)
     image_push(image) if push
   end
 
+  def fetch_image(image)
```
justin808, line 61: rename to `fetch_image_details`

lib/core/controlplane_api.rb
```diff
@@ -25,6 +25,10 @@ def query_images(org:, gvc:, gvc_op_type:)
     query("/org/#{org}/image", terms)
   end
 
+  def fetch_image(org:, image:)
```
justin808, line 28: rename to `fetch_image_details`

justin808 (changes requested):

PR #159, opened 2024-05-02
Updates templates naming

rafaelgomesxyz:
- Renames `gvc` templates to `app`
- Adds `identity` inside `app` templates
- Adds `secrets-policy` inside `secrets` templates
- Adds option to set custom names for `secrets` and `secrets-policy`


spec/dummy/.controlplane/templates/secrets.yml
```diff
@@ -1,3 +1,8 @@
+kind: secret
```
rafaelgomesxyz, line 1: Decided to keep it named `secrets.yml` instead of `org.yml` because:

1. `org.yml` would not create an org, so it would be misleading
2. You could apply `secrets.yml` multiple times if using different secrets for different apps (for example, different secrets for staging and review apps in the same org)

justin808: I disagree with point "1." Yes, we don't create an org, but this is a one time thing for the org.

I don't understand point number two. Maybe you're trying to say in case you have multiple orgs, then you need to apply this to multiple orgs?

In any case, I'd prefer if this is org.yml. Yes, the org is set up in the UX. So what? After that, you apply the org.yml file. I found the top part 

https://github.com/shakacode/heroku-to-control-plane/pull/159/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R200-R201

```
    # If you're going to use secrets, you need to apply the `secrets.yml` template separately (one-time setup):
    # `cpl apply-template secrets -a my-app`
``

very confusing...

```
    # If you're going to use secrets, you need to apply the `org.yml` template (one-time setup for your org):
    # Pick any app in the relevant org.
    # `cpl apply-template org -a my-app`
```

CC: @dzirtusss

File: README.md [hunk 2]
```diff
@@ -189,27 +189,40 @@ aliases:
...
+    # instead of `cpl apply-template app redis postgres memcached rails sidekiq`.
     #
     # Note:
     # 1. These names correspond to files in the `./controlplane/templates` directory.
     # 2. Each file can contain many objects, such as in the case of templates that create a resource, like `postgres`.
     # 3. While the naming often corresponds to a workload or other object name, the naming is arbitrary. 
     #    Naming does not need to match anything other than the file name without the `.yml` extension.
+    #
+    # If you're going to use secrets, you need to apply the `secrets.yml` template separately (one-time setup):
+    # `cpl apply-template secrets -a my-app`
```
justin808, line 201: See the comment at the bottom regarding naming this file `org.yml`.

It's a one time thing per org.

CC: @dzirtusss

README.md
```diff
@@ -143,7 +143,7 @@ The `cpl` gem is based on several configuration files within a `/.controlplane`
 ```
 .controlplane/
 ├─ templates/
-│  ├─ gvc.yml
+│  ├─ app.yml
 │  ├─ postgres.yml
 │  ├─ rails.yml
```
justin808, line 148: add `org.yml` here

README.md
```diff
@@ -154,8 +154,8 @@ The `cpl` gem is based on several configuration files within a `/.controlplane`
 1. `controlplane.yml` describes the overall application. Be sure to have `<your-org>` as the value for `aliases.common.cpln_org`, or set it with the `CPLN_ORG` environment variable.
 2. `Dockerfile` builds the production application. `entrypoint.sh` is an _example_ entrypoint script for the production application, referenced in your Dockerfile.
 3. `templates` directory contains the templates for the various workloads, such as `rails.yml` and `postgres.yml`.
-4. `templates/gvc.yml` defines your project's GVC (like a Heroku app). More importantly, it contains ENV values for the app.
-5. `templates/rails.yml` defines your Rails workload. It may inherit ENV values from the parent GVC, which is populated from the `templates/gvc.yml`. This file also configures scaling, sizing, firewalls, and other workload-specific values.
+4. `templates/app.yml` defines your project's GVC (like a Heroku app). More importantly, it contains ENV values for the app.
+5. `templates/rails.yml` defines your Rails workload. It may inherit ENV values from the parent GVC, which is populated from the `templates/app.yml`. This file also configures scaling, sizing, firewalls, and other workload-specific values.
```
justin808, line 158: add `org.yml` and explain.

justin808 (changes requested):
1. rename `secrets.yml` to `org.yml`
2. better explanation of secrets and `org.yml` in docs.

spec/dummy/.controlplane/templates/secrets.yml
justin808: Sergey says this is not for staging and production

justin808: OK -- I get this now. the app name is used to create the right secrets...

and if you specify the command multiple times with different app names, then different secrets get created.

I suggest that we automatically create the org secrets when `setup-app` is used and the secrets and policy don't exist (unless `--skip-secret-access-binding`) is used. 

```
kind: secret
name: {{APP_SECRETS}}
type: dictionary
data: {}
---
# Policy is needed to allow identities to access secrets
kind: policy
name: {{APP_SECRETS_POLICY}}
targetKind: secret
targetLinks:
  - //secret/{{APP_SECRETS}}
```

justin808: We can do this in a follow up PR.

PR #163, opened 2024-05-10
Fix race conditions in `run` command

rafaelgomesxyz:
Fixes race conditions in `run` command by waiting for the exact expected deployed version.

## Summary by CodeRabbit

- **New Features**
	- Added a method to retrieve the last processed version of workload deployments when ready.
- **Enhancements**
	- Improved the initialization and dynamic update capabilities of runner workloads, including CPU and memory adjustments.
- **Bug Fixes**
	- Enhanced logic for workload version increments and conditions for workload creation and updates.
- **Refactor**
	- Removed unnecessary complexity in code, improving maintainability.

### DIFF
--- a/lib/command/run.rb
+++ b/lib/command/run.rb
@@ -87,9 +87,9 @@ module Command
     MAGIC_END = "---cpl run command finished---"
 
     attr_reader :interactive, :detached, :location, :original_workload, :runner_workload,
-                :container, :image_link, :image_changed, :job, :replica, :command
+                :container, :expected_deployed_version, :job, :replica, :command
 
-    def call # rubocop:disable Metrics/CyclomaticComplexity, Metrics/MethodLength, Metrics/PerceivedComplexity
+    def call # rubocop:disable Metrics/MethodLength
       @interactive = config.options[:interactive] || interactive_command?
       @detached = config.options[:detached]
       @log_method = config.options[:log_method]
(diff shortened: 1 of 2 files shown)


lib/command/run.rb
```diff
@@ -177,41 +172,54 @@ def create_runner_workload # rubocop:disable Metrics/MethodLength
...
-        cp.fetch_workload(runner_workload)
+        !cp.cron_workload_deployed_version(runner_workload).nil?
       end
     end
 
     def wait_for_runner_workload_update
       step("Waiting for runner workload '#{runner_workload}' to be updated", retry_on_failure: true) do
-        _, container_spec = base_workload_specs(runner_workload)
-        container_spec["image"] == image_link
+        cp.cron_workload_deployed_version(runner_workload) == expected_deployed_version
```
rafaelgomesxyz, line 222: This could be an endless loop if the workload is updated by someone else at the same time or if another `run` job is starting at the same time, because then the version would increase. Perhaps we should use `>=` instead of `==`. Then again, that wouldn't fix the issue of simultaneous jobs.

justin808: Yes, 100%.

And log the situation. 

Maybe we're better off with the temp workload strategy b/c of the race condition?

justin808:
```suggestion
        cp.cron_workload_deployed_version(runner_workload) >= expected_deployed_version
```

lib/command/run.rb
```diff
@@ -177,41 +172,54 @@
...
+        end
+        if config.options[:memory] && container_spec["memory"] != config.options[:memory]
+          container_spec["memory"] = config.options[:memory]
+          should_update = true
+        end
+
+        next true unless should_update
 
         # Update runner workload
+        @expected_deployed_version += 1
```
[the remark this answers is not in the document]
justin808: what is the difference of your suggestion?

PR #167, opened 2024-05-13
Raise error if org does not exist

### LINKED ISSUE #166: Raise error if org does not exist
When making requests to Control Plane, we should check if the org exists and show an error if not.

rafaelgomesxyz:
Fixes #166

## Summary by CodeRabbit

- **New Features**
	- Added a method to verify the existence of an organization before proceeding with commands.
	- Introduced a new method to list all organizations.
- **Bug Fixes**
	- Improved error handling in the `deploy-image` command.
- **Refactor**
	- Updated the `run` command behavior and deprecated the `run:detached` command.
- **Tests**
	- Added tests to ensure organizational checks are in place during initialization and command execution.

### DIFF
--- a/lib/core/controlplane_api.rb
+++ b/lib/core/controlplane_api.rb
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
 class ControlplaneApi # rubocop:disable Metrics/ClassLength
+  def list_orgs
+    api_json("/org", method: :get)
+  end
+
   def gvc_list(org:)
     api_json("/org/#{org}/gvc", method: :get)
   end
--- a/spec/core/controlplane_spec.rb
+++ b/spec/core/controlplane_spec.rb
@@ -3,10 +3,19 @@
 require "spec_helper"
 
 describe Controlplane do
-  let!(:fake_config) { Struct.new(:app, :org).new("my-app", "my-org") }
-  let!(:described_instance) { described_class.new(fake_config) }
+  describe "#initialize" do
+    let!(:fake_config) { Struct.new(:app, :org).new("my-app", "my-org") }
+
+    it "raises error if org does not exist" do
+      expect do
+        described_class.new(fake_config)
+      end.to raise_error(include("Can't find org 'my-org'"))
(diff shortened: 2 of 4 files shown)


lib/core/controlplane_api.rb
```diff
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
 class ControlplaneApi # rubocop:disable Metrics/ClassLength
+  def list_orgs
```
dzirtusss, line 4: Style, I guess `org_list` is more appropriate name (noun vs verb). Plus it will be the same style `gvc_list` below.

spec/core/controlplane_spec.rb
```diff
@@ -3,10 +3,19 @@
 require "spec_helper"
 
 describe Controlplane do
-  let!(:fake_config) { Struct.new(:app, :org).new("my-app", "my-org") }
-  let!(:described_instance) { described_class.new(fake_config) }
+  describe "#initialize" do
+    let!(:fake_config) { Struct.new(:app, :org).new("my-app", "my-org") }
```
dzirtusss, line 7: Overall, `let!` with bang is a bad style, especially you "overuse" it. Instead better normal `let` or local var or a method (whatever applicable).

I mean for future, not this PR, try avoiding `let!`

dzirtusss (approved):

lib/core/controlplane_api.rb
rafaelgomesxyz: We need to reach an agreement on what to use, because you prefer "noun + verb", but Justin prefers "verb + noun", so right now all of the method names are very inconsistent. Let's decide on a strategy and apply it in a separate PR.

spec/core/controlplane_spec.rb
rafaelgomesxyz: Ok, I was using it because for `let!(:app)` because the app needs to be setup before the test begins, so we can't let it be lazily evaluated, but I guess we don't need to use it everywhere.

PR #170, opened 2024-05-13
Unbind identity from policy when deleting app in `delete` command

rafaelgomesxyz:
We do not currently clean policy bindings, which can cause issues, as the maximum number of bindings is 200. Once that limit is reached, new bindings will fail to be created.

This PR fixes that by unbinding the identity from the policy when deleting an app in the `delete` command. However, it does not work retroactively (old bindings will have to be cleaned manually).

## Summary by CodeRabbit

- **New Features**
  - Enhanced `delete` command to unbind the app from the secrets policy if both the identity and policy exist and are bound.

- **Documentation**
  - Updated command documentation to reflect the new unbinding functionality for the `delete` command.

- **Tests**
  - Added new test scenarios to verify the unbinding of identity from policy under various conditions.

- **Configuration**
  - Introduced a new dummy configuration for testing without a secrets policy.


dzirtusss (approved):
LGTM

PR #177, opened 2024-05-17
Fix job exit logic

dzirtusss:
- Fix job exit logic not handling all the cases, especially "active" when job is still running.
- Fix magic string appearing in interactive mode (it should not) and proper exit status.

## Summary by CodeRabbit

- **New Features**
  - Enhanced job status handling to continuously check and update the status, improving real-time feedback for job execution.

- **Tests**
  - Added new test cases to verify job status handling, including scenarios for "successful," "failed," and "active" statuses.

### DIFF
--- a/lib/command/run.rb
+++ b/lib/command/run.rb
@@ -395,7 +395,7 @@ module Command
       script += interactive_runner_script if interactive
 
       script +=
-        if @log_method == 1
+        if @log_method == 1 || @interactive
           args_join(config.args)
         else
           <<~SCRIPT
--- a/lib/command/run.rb
+++ b/lib/command/run.rb
@@ -448,12 +448,11 @@ module Command
         job_details = result&.dig("items")&.find { |item| item["id"] == job }
         status = job_details&.dig("status")
 
-        progress.puts("Job status: #{status}")
+        Shell.debug("Job status: #{status}")
 
         case status
         when "active"
           sleep 1
-          redo
         when "successful"
           break ExitCode::SUCCESS
         else
--- a/lib/command/run.rb
+++ b/lib/command/run.rb
@@ -442,16 +442,23 @@ module Command
       )
     end
 
-    def resolve_job_status
-      result = cp.fetch_cron_workload(runner_workload, location: location)
-      job_details = result&.dig("items")&.find { |item| item["id"] == job }
-      status = job_details&.dig("status")
+    def resolve_job_status # rubocop:disable Metrics/MethodLength
+      loop do
+        result = cp.fetch_cron_workload(runner_workload, location: location)
+        job_details = result&.dig("items")&.find { |item| item["id"] == job }
+        status = job_details&.dig("status")
 
-      case status
-      when "failed"
-        ExitCode::ERROR_DEFAULT
-      when "successful"
-        ExitCode::SUCCESS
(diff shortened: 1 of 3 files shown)


lib/command/run.rb
```diff
@@ -395,7 +395,7 @@ def runner_script # rubocop:disable Metrics/MethodLength
       script += interactive_runner_script if interactive
 
       script +=
-        if @log_method == 1
+        if @log_method == 1 || @interactive
```
justin808, line 398: @dzirtusss 1 is a magic number. Why not some enum?

lib/command/run.rb
```diff
@@ -442,16 +442,22 @@ def print_detached_commands
       )
     end
 
-    def resolve_job_status
-      result = cp.fetch_cron_workload(runner_workload, location: location)
-      job_details = result&.dig("items")&.find { |item| item["id"] == job }
-      status = job_details&.dig("status")
+    def resolve_job_status # rubocop:disable Metrics/MethodLength
```
justin808, line 445: rename

`wait_for_job_completion`

and add some method documentation

lib/command/run.rb
```diff
@@ -442,16 +442,22 @@ def print_detached_commands
...
+        Shell.debug("JOB STATUS", status)
 
-      case status
-      when "failed"
-        ExitCode::ERROR_DEFAULT
-      when "successful"
-        ExitCode::SUCCESS
+        case status
+        when "active"
+          sleep 1
```
justin808, line 455: Should this be a constant defined at the top of the file?

```ruby
JOB_COMPLETION_POLLING_PERIOD = 1
```

lib/command/run.rb
```diff
@@ -442,16 +442,22 @@ def print_detached_commands
       )
     end
 
-    def resolve_job_status
-      result = cp.fetch_cron_workload(runner_workload, location: location)
-      job_details = result&.dig("items")&.find { |item| item["id"] == job }
-      status = job_details&.dig("status")
+    def resolve_job_status # rubocop:disable Metrics/MethodLength
+      loop do
```
justin808, line 446: @dzirtusss should we have some catch if this loop is going to be infinite, like the workload got renamed so `result` is always nil?

justin808: https://github.com/shakacode/control-plane-flow/blob/fe6493f86956b1e2e89ff140abeb96b5391faec9/lib/core/controlplane.rb#L245C1-L248C6

```ruby
  def fetch_cron_workload(workload, location:)
    cmd = "cpln workload cron get #{workload} #{gvc_org} --location #{location} -o yaml"
    perform_yaml(cmd)
  end
```

`perform_yaml` should always add the `-o yaml`

lib/command/run.rb
```diff
@@ -442,16 +442,22 @@ def print_detached_commands
...
+        result = cp.fetch_cron_workload(runner_workload, location: location)
```
justin808, line 447: This could return nil for 1,000 reasons. Can we make this more robust?

Should we generally use the version of `perform_yaml!` that exits if non-zero?

In general we should crash immediately for unexpected output.

If we expect that perform_yaml might have a non-zero exit code, then check std_err for the expected error and otherwise abort.

justin808 (approved):
Approved, but needs FAST followups.

PR #178, opened 2024-05-17
Disable interrupt trap in tests

rafaelgomesxyz:
Currently, pressing `Ctrl + C` while running the tests does not stop them, because we trap the interrupt signal in the app, which conflicts with RSpec's trap.

This PR adds a new flag `DISABLE_INTERRUPT_TRAP` to prevent trapping the interrupt signal, which we can use in the tests to make sure that `Ctrl + C` works properly.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new method to handle interrupt signals more gracefully, ensuring a smoother exit process.
  
- **Bug Fixes**
  - Improved interrupt handling to prevent abrupt terminations during operations.

- **Tests**
  - Added a new environment variable setting for testing interrupt handling scenarios.

### DIFF
--- a/lib/core/shell.rb
+++ b/lib/core/shell.rb
@@ -90,4 +90,11 @@ class Shell
 
     message.gsub(pattern, "XXXXXXX")
   end
+
+  def self.trap_interrupt
+    trap("SIGINT") do
+      puts
+      exit(ExitCode::INTERRUPT)
+    end
+  end
 end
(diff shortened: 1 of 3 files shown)


lib/core/shell.rb
```diff
@@ -90,4 +90,11 @@ def self.hide_sensitive_data(message, pattern = nil)
 
     message.gsub(pattern, "XXXXXXX")
   end
+
+  def self.trap_interrupt
```
rafaelgomesxyz, line 94: Just moved here from `cpl.rb`.

PR #179, opened 2024-05-23
Add global identifier for dummy test apps

rafaelgomesxyz:
We cannot do multiple test runs at the same time currently, because the specs are not designed to run in parallel.

This PR changes that, by adding a global identifier at the end of the prefix for the app names, ensuring that, while some apps may have the same name in the same test run (for the purpose of saving time), they'll have different names in different test runs.

We also reduce the size of randomly generated suffixes from 8 to 4, and rename the apps so that the names are shorter, because Control Plane has a limit of 64 characters.

## Summary by CodeRabbit

- **Chores**
  - Updated test app names for better consistency and clarity across various test scenarios.
  - Modified `.gitignore` to exclude specific temporary files.
  - Renamed app configurations in controlplane settings to include dynamic identifiers for improved manageability.


spec/command/apply_template_spec.rb
```diff
@@ -135,7 +135,7 @@
   end
 
   context "when workload already exists" do
-    let!(:app) { dummy_test_app("with-rails", create_if_not_exists: true) }
+    let!(:app) { dummy_test_app("rails", create_if_not_exists: true) }
```
rafaelgomesxyz, line 138: Renamed a lot of the apps so that the names are shorter, because Control Plane has a limit of 64 characters.

spec/command/copy_image_from_upstream_spec.rb
```diff
@@ -7,6 +7,7 @@
     let!(:app) { dummy_test_app }
 
     before do
+      allow(Shell).to receive(:cmd).and_call_original
```
rafaelgomesxyz, line 10: Specs were failing locally without this.

spec/command/setup_app_spec.rb
```diff
@@ -78,23 +78,4 @@
       expect(result[:stderr]).to match(/Binding identity to policy[.]+? done!/)
     end
   end
-
-  context "when using custom names for secrets" do
```
rafaelgomesxyz, line 82: The spec above ("binds identity to policy") now uses custom names, because with the global identifier we don't want to test using default names, as that will create a different secret/policy for each global identifier.

spec/dummy/.controlplane/controlplane.yml
```diff
@@ -11,7 +11,7 @@ aliases:
       - postgres
 
 apps:
-  dummy-test:
+  dummy-test-{GLOBAL_IDENTIFIER}:
```
rafaelgomesxyz, line 14: We now need to add this placeholder and replace it when running the tests to ensure that everything works properly. We replace it in `dummy_app_setup.rb`.

spec/support/command_helpers.rb
```diff
@@ -64,7 +69,7 @@ def dummy_test_app_prefix(extra_prefix = "")
   # Returns the app name.
   def dummy_test_app(extra_prefix = "", suffix = "", create_if_not_exists: false) # rubocop:disable Metrics/CyclomaticComplexity
     prefix = dummy_test_app_prefix(extra_prefix)
-    suffix = SecureRandom.hex(4) if (suffix.nil? || suffix.empty?) && !create_if_not_exists
+    suffix = SecureRandom.hex(2) if (suffix.nil? || suffix.empty?) && !create_if_not_exists
```
rafaelgomesxyz, line 72: Reduced the size so that the names are shorter, because Control Plane has a limit of 64 characters.

It should still be rare to run into conflicts, so I don't think we need to worry about that for now.

spec/support/command_helpers.rb
```diff
@@ -38,13 +38,18 @@ def dummy_test_org
...
+  end
+
   # `extra_prefix` is used to differentiate between different dummy apps,
   # e.g., "dummy-test-default", "dummy-test-with-nothing", etc.
   #
   # Returns the full prefix.
   def dummy_test_app_prefix(extra_prefix = "")
     prefix = DUMMY_TEST_APP_PREFIX
     prefix += "-#{extra_prefix}" unless extra_prefix.nil? || extra_prefix.empty?
+    prefix += "-#{dummy_test_app_global_identifier}"
```
rafaelgomesxyz, line 52: We now add a global identifier at the end of the prefix, and since `@@global_identifier` is a class var, that ensures that the apps will have different names in different test runs.

spec/support/dummy_app_setup.rb
```diff
@@ -4,19 +4,29 @@ module DummyAppSetup
...
+    config_dir = "#{CommandHelpers.spec_directory}/dummy/.controlplane"
+
+    config_file = File.read("#{config_dir}/controlplane.yml")
+    config_file = config_file.gsub("{GLOBAL_IDENTIFIER}", CommandHelpers.dummy_test_app_global_identifier)
+
+    @tmp_config_file = Tempfile.create(["controlplane-tmp-", ".yml"], config_dir)
+    @tmp_config_file.write(config_file)
+    @tmp_config_file.rewind
+
+    ENV["CONFIG_FILE_PATH"] = @tmp_config_file.path
```
rafaelgomesxyz, line 16: We need to read the original `controlplane.yml` file, replace the `{GLOBAL_IDENTIFIER}` placeholder, and save it to a new temporary file in the same directory.

spec/support/dummy_app_setup.rb
```diff
@@ -4,19 +4,29 @@ module DummyAppSetup
...
+    else
+      CommandHelpers.apps_to_delete.each do |app|
+        CommandHelpers.run_cpl_command("delete", "-a", app, "--yes")
+      end
     end
 
-    CommandHelpers.apps_to_delete.each do |app|
-      CommandHelpers.run_cpl_command("delete", "-a", app, "--yes")
-    end
+    File.delete(@tmp_config_file.path) if @tmp_config_file
```
rafaelgomesxyz, line 30: We should delete the temporary config file at the end of the run.

.gitignore
```diff
@@ -15,3 +15,4 @@
 .rspec_status
 
 /spec.log
+/spec/dummy/.controlplane/controlplane-tmp-*.yml
```
rafaelgomesxyz, line 18: Ignore any temporary config files, in case the cleanup in `dummy_app_setup.rb` doesn't work.

PR #182, opened 2024-05-24
Update `run` command based on new `cpln` changes

### LINKED ISSUE #164: Race conditions with run command
The goal is to offer something like heroku run ([docs](https://devcenter.heroku.com/articles/heroku-cli-commands#heroku-run)) and heroku run:detached ([docs](https://devcenter.heroku.com/articles/heroku-cli-commands#heroku-run-detached)).

We’ve run into a race condition, and we’d like some advice.

[@Rafael Gomes](https://shakacode.slack.com/team/U032Q422LG4)
 tried to address this in [PR 163](https://github.com/shakacode/heroku-to-control-plane/pull/163).

# Our current strategy:
1. Clone the main application standard workload and make a cron workload with as many settings the same as feasible. The new workload name gets the original name plus -runner
2. When running a job, update the cron workload definition if needed, on the following criteria ([code here](https://github.com/shakacode/heroku-to-control-plane/blob/main/lib/command/run.rb#L175-L212)):
  a. **Image**: Sometimes we need to use a newer image than is the deployed image, such as during a release, where we have new code that will have new database migrations.
  b. **CPU** and **Memory**

Thus, we could have a race condition if we’re updating the workload, and we try to kick off several jobs with different settings.  
While it would be unfortunate to run some code with the wrong memory or CPU, running with the wrong image would be totally unacceptable.

Previously, we didn’t have this issue as we always made a new workload for each job run. I’m guessing that it’s much cleaner to have one workload configured for these run commands. However, if we need to change the workload and then start a job, there’s a big race condition issue.

I see a couple ways out of this:

1. We create multiple workloads for job runners, with a name based on the image version, and maybe the CPU and memory. Tricky thing here is when to clean these up when not used. For example, if we need to run a job with the latest image, we will use a workload runner called something like `rails-runner-latest`.  The `rails-runner` workload would ALWAYS have the same image as the `rails` workload.
2. CPLN provides updates the cpln workload cron start command to allow setting the image, cpu, and ram.

rafaelgomesxyz:
Fixes #164

`cpln` now accepts overriding `image`, `cpu`, and `memory` when starting a Cron workload job, so we no longer have to worry about race conditions when running simultaneous jobs with different overrides.

This PR updates the `run` command to reflect those changes.

Also changes jobs to use a CPU size of 1 (1 core) and a memory size of 2Gi (2 gibibytes) by default (configurable through `runner_job_default_cpu` and `runner_job_default_memory` in `controlplane.yml`), as just copying the sizes from the original workload may lead to an overprovision of resources.

## Summary by CodeRabbit

- **New Features**
  - Added ability to override `--image`, `--cpu`, and `--memory` for each job individually in the `run` command.
  - Introduced default CPU and memory sizes for jobs (1 core and 2 GiB, respectively).

- **Improvements**
  - Specific validations now run before commands, exiting with a non-zero code if validation fails.
  - Replaced deprecated `--skip-secret-access-binding` option with `--skip-secrets-setup`.
  - Terminal size synchronization between local and remote terminals can now be configured.

- **Documentation**
  - Updated documentation to reflect new default CPU and memory settings and the ability to override these settings per job.
  - Clarified usage of `--entrypoint` and terminal size configuration options.

### DIFF
--- a/lib/command/run.rb
+++ b/lib/command/run.rb
@@ -434,6 +434,8 @@ module Command
     end
 
     def print_detached_commands
+      return unless replica
+
       app_workload_replica_config = app_workload_replica_args.join(" ")
       progress.puts(
         "\n\n" \
--- a/lib/command/run.rb
+++ b/lib/command/run.rb
@@ -103,7 +103,7 @@ module Command
       @location = config.location
       @original_workload = config.options[:workload] || config[:one_off_workload]
       @runner_workload = "#{original_workload}-runner"
-      @default_image = cp.latest_image_from([], app_name: config.app)
+      @default_image = "#{config.app}:#{Controlplane::NO_IMAGE_AVAILABLE}"
       @default_cpu = config.current[:runner_job_default_cpu] || DEFAULT_JOB_CPU
       @default_memory = config.current[:runner_job_default_memory] || DEFAULT_JOB_MEMORY
 
--- a/lib/command/run.rb
+++ b/lib/command/run.rb
@@ -451,7 +451,7 @@ module Command
         Shell.debug("JOB STATUS", status)
 
         case status
-        when "active"
+        when "active", "pending"
           sleep 1
         when "successful"
           break ExitCode::SUCCESS
--- a/lib/command/run.rb
+++ b/lib/command/run.rb
@@ -36,12 +36,15 @@ module Command
       - - log async fetching for non-interactive mode
       - The Dockerfile entrypoint is used as the command by default, which assumes `exec "${@}"` to be present,
         and the args ["bash", "-c", cmd_to_run] are passed
-      - The entrypoint can be overriden through `--entrypoint`, which must be a single command or a script path that exists in the container,
+      - The entrypoint can be overridden through `--entrypoint`, which must be a single command or a script path that exists in the container,
         and the args ["bash", "-c", cmd_to_run] are passed,
         unless the entrypoint is `bash`, in which case the args ["-c", cmd_to_run] are passed
       - Providing `--entrypoint none` sets the entrypoint to `bash` by default
       - If `fix_terminal_size` is `true` in the `.controlplane/controlplane.yml` file,
-        the remote terminal size will be fixed to match the local terminal size (may also be overriden through `--terminal-size`)
+        the remote terminal size will be fixed to match the local terminal size (may also be overridden through `--terminal-size`)
+      - By default, all jobs use a CPU size of 1 (1 core) and a memory size of 2Gi (2 gibibytes)
+        (can be configured through `runner_job_default_cpu` and `runner_job_default_memory` in `controlplane.yml`,
+        and also overridden per job through `--cpu` and `--memory`)
     DESC
     EXAMPLES = <<~EX
       ```sh
@@ -84,12 +87,15 @@ module Command
       ```
     EX
 
+    DEFAULT_JOB_CPU = "1"
+    DEFAULT_JOB_MEMORY = "2Gi"
     MAGIC_END = "---cpl run command finished---"
 
     attr_reader :interactive, :detached, :location, :original_workload, :runner_workload,
+                :default_image, :default_cpu, :default_memory,
                 :container, :expected_deployed_version, :job, :replica, :command
 
-    def call # rubocop:disable Metrics/MethodLength
+    def call # rubocop:disable Metrics/CyclomaticComplexity, Metrics/MethodLength, Metrics/PerceivedComplexity
       @interactive = config.options[:interactive] || interactive_command?
       @detached = config.options[:detached]
       @log_method = config.options[:log_method]
@@ -97,6 +103,9 @@ module Command
       @location = config.location
       @original_workload = config.options[:workload] || config[:one_off_workload]
       @runner_workload = "#{original_workload}-runner"
+      @default_image = cp.latest_image_from([], app_name: config.app)
+      @default_cpu = config.current[:runner_job_default_cpu] || DEFAULT_JOB_CPU
+      @default_memory = config.current[:runner_job_default_memory] || DEFAULT_JOB_MEMORY
 
       unless interactive
         @internal_sigint = false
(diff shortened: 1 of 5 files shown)


lib/command/run.rb
```diff
@@ -172,57 +170,12 @@ def create_runner_workload # rubocop:disable Metrics/MethodLength
...
-
-        # Container overrides
-        if config.options[:cpu] && container_spec["cpu"] != config.options[:cpu]
-          container_spec["cpu"] = config.options[:cpu]
-          should_update = true
-        end
-        if config.options[:memory] && container_spec["memory"] != config.options[:memory]
-          container_spec["memory"] = config.options[:memory]
-          should_update = true
-        end
```
rafaelgomesxyz, line 204: The overrides have been moved to `build_start_job_yaml` below.

lib/command/run.rb
```diff
@@ -308,6 +261,8 @@ def base_workload_specs(workload)
     end
 
     def build_job_start_yaml # rubocop:disable Metrics/CyclomaticComplexity, Metrics/MethodLength, Metrics/PerceivedComplexity
+      _, container_spec = base_workload_specs(original_workload)
```
rafaelgomesxyz, line 264: Even though `container_spec` is only used in line 298, we need to call `base_workload_specs` here because it sets `@container`.

lib/command/run.rb
```diff
@@ -335,6 +290,18 @@ def build_job_start_yaml # rubocop:disable Metrics/CyclomaticComplexity, Metrics
         job_start_hash["args"].push('eval "$CPL_RUNNER_SCRIPT"')
       end
 
+      image = config.options[:image]
+      image_link = if image
+                     image = latest_image if image == "latest"
+                     "/org/#{config.org}/image/#{image}"
+                   else
+                     container_spec["image"]
```
justin808, line 298: What is the purpose of setting this if it's the same? Some race condition?

lib/command/run.rb
justin808: sounds very hacky, and not sure that container_spec is needed.

lib/command/run.rb
rafaelgomesxyz: Renamed it to `original_container_spec` so it's clearer: 84361d349c7c9d55acb924772e3d427df9da5132

lib/command/run.rb
rafaelgomesxyz: We do need it. See if 1ce60c4db193410852977c67ce4f76aaa074ee74 makes it better. I moved the `@container` assignment out of `base_workload_specs`.

lib/command/run.rb
```diff
@@ -434,6 +433,8 @@ def wait_for_job_status_and_log(logs_pipe) # rubocop:disable Metrics/MethodLengt
     end
 
     def print_detached_commands
+      return unless replica
```
rafaelgomesxyz, line 436: Just to prevent printing the detached commands if pressing `Ctrl + C` too early (before the replica is created).

lib/command/run.rb
```diff
@@ -451,7 +452,7 @@ def resolve_job_status # rubocop:disable Metrics/MethodLength
         Shell.debug("JOB STATUS", status)
 
         case status
-        when "active"
+        when "active", "pending"
```
rafaelgomesxyz, line 455: Some jobs may have the status as `pending` before moving to `active`, so we need to wait for those as well.

lib/command/run.rb
```diff
@@ -84,19 +87,25 @@ class Run < Base # rubocop:disable Metrics/ClassLength
...
-    def call # rubocop:disable Metrics/MethodLength
+    def call # rubocop:disable Metrics/CyclomaticComplexity, Metrics/MethodLength, Metrics/PerceivedComplexity
       @interactive = config.options[:interactive] || interactive_command?
       @detached = config.options[:detached]
       @log_method = config.options[:log_method]
 
       @location = config.location
       @original_workload = config.options[:workload] || config[:one_off_workload]
       @runner_workload = "#{original_workload}-runner"
+      @default_image = cp.latest_image_from([], app_name: config.app)
```
dzirtusss, line 106: I guess default image should be the current image of the app and not the latest? I mean if we want latest we do `--latest` which is easy. How you do the current one? `--current`? I mean surely one can specify exact, but this is complex.

Anyway this may be a breaking change.

lib/command/run.rb
dzirtusss: 👍

dzirtusss (changes requested):

lib/command/run.rb
rafaelgomesxyz: This is actually not the latest image, and it has nothing to do with which image is used by the jobs. Passing an empty array to `cp.latest_image_from` simply returns `{APP_NAME}:NO_IMAGE_AVAILABLE`:

https://github.com/shakacode/control-plane-flow/blob/2133ff4be886d978b1becec7b7f9ae8c4943fb4a/lib/core/controlplane.rb#L65-L71

I just wanted to reset the runner workload image, because we won't be updating it anymore. If we were to set it to anything else, it would eventually become outdated, so that's why I thought `NO_IMAGE_AVAILABLE` would be better.

Images are now set for each job separately, in `build_job_start_yaml`. And there, we have the correct logic for defaulting to the current image:

https://github.com/shakacode/control-plane-flow/blob/2133ff4be886d978b1becec7b7f9ae8c4943fb4a/lib/command/run.rb#L345-L353

Perhaps I can rename the `@default_image` variable to make it less confusing? It's meant to be "default image for workload", not "default image for jobs". So maybe `@workload_default_image`?

dzirtusss: Yes, `latest` word is confusing as it has different meaning as term (please change it). Overall, I guess I more like `NO_IMAGE` way for the runner. And set image per job.

rafaelgomesxyz: Ok, done in 1c6e103baff6c972e3a2db9b574552f6e23fd27d.

dzirtusss (approved):
LGTM

justin808 (approved):

PR #211, opened 2024-07-18
Separate docker build & docker push steps

zzaakiirr:
### What does this PR?

This PR separates `docker build` and `docker push` steps

### Why?

1. Don't run `cpln image docker-login --org <org_name>` so that push to private registry (`<org_name>.registry.cpln.io`) won't succeed
2. Run `cpflow build-image -a <app_name>`

This will output something like this:
```
Building image from Dockerfile '<path_to_dockerfile>'...
ERROR: Command exited with non-zero status.
```

This message is misleading because `docker build` succeeds but `docker push` fails

## Summary by CodeRabbit

- **New Features**
  - Enhanced image build process to automatically push images to a specified path with progress updates.
- **Improvements**
  - Simplified the image build method to focus solely on the Docker image construction, removing unnecessary push functionality for clearer operation.

### DIFF
--- a/lib/command/build_image.rb
+++ b/lib/command/build_image.rb
@@ -36,8 +36,7 @@ module Command
 
       cp.image_build(image_url, dockerfile: dockerfile,
                                 docker_args: config.args,
-                                build_args: build_args,
-                                push: false)
+                                build_args: build_args)
 
       push_path = "/org/#{config.org}/image/#{image_name}"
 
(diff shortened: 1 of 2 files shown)


lib/command/build_image.rb
```diff
@@ -36,9 +36,14 @@ def call # rubocop:disable Metrics/MethodLength
 
       cp.image_build(image_url, dockerfile: dockerfile,
                                 docker_args: config.args,
-                                build_args: build_args)
+                                build_args: build_args,
+                                push: false)
```
rafaelgomesxyz, line 40: We can remove `push` from `image_build`, since it was only being used here. I can't think of another scenario where it would be used.

rafaelgomesxyz (changes requested):

rafaelgomesxyz (approved):
LGTM

PR #219, opened 2024-07-25
Remove config validation for cpflow generate command

zzaakiirr:
### What does this PR do?

This PR:
- removes `config` validation from `cpflow generate` command 
- fixes specs for this command

## Summary by CodeRabbit

- **New Features**
	- Introduced a dedicated constant for validations in the `Generate` class, allowing for future enhancement of validation rules.

- **Bug Fixes**
	- Improved test robustness by adding metadata tags for validation scenarios and streamlined configuration file checks in tests.

- **Documentation**
	- Enhanced clarity in test setups by defining configuration file paths and ensuring isolated testing conditions.

- **Refactor**
	- Consolidated configuration file handling into a new method for improved readability and maintainability.

### DIFF
--- a/spec/command/generate_spec.rb
+++ b/spec/command/generate_spec.rb
@@ -17,7 +17,9 @@ ensure
   Dir.chdir original_working_dir
 end
 
-describe Command::Generate do
+describe Command::Generate, :enable_validations, :without_config_file do
+  let(:controlplane_config_file_path) { CONTROLPLANE_CONFIG_DIR_PATH.join("controlplane.yml") }
+
   before do
     FileUtils.rm_r(GENERATOR_PLAYGROUND_PATH) if Dir.exist?(GENERATOR_PLAYGROUND_PATH)
     FileUtils.mkdir_p GENERATOR_PLAYGROUND_PATH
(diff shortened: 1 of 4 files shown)


zzaakiirr:
@coderabbitai review

borela (approved):

spec/command/generate_spec.rb
```diff
@@ -17,7 +17,9 @@ def inside_dir(path)
   Dir.chdir original_working_dir
 end
 
-describe Command::Generate do
+describe Command::Generate, :enable_validations, :without_config_file do
```
rafaelgomesxyz, line 20: Why do we enable validations?

spec/spec_helper.rb
```diff
@@ -152,4 +152,16 @@
     # Times out after 10 minutes
     Timeout.timeout(600) { example.run }
   end
+
+  config.around(:example, :without_config_file) do |example|
+    CommandHelpers.unset_config_file
```
rafaelgomesxyz, line 157: This doesn't delete the tempfile created by `configure_config_file`. Maybe we should just call `delete_config_file` here and set `ENV["CONFIG_FILE_PATH"]` to `nil` in there, without the need for `unset_config_file`.

rafaelgomesxyz (changes requested):

spec/command/generate_spec.rb
zzaakiirr: We need to check real case scenario (where validations are enabled).

Otherwise test case for `"when no configuration exist in the project"` context passes even if we don't override `Command::Generate::VALIDATIONS` constant

spec/spec_helper.rb
rafaelgomesxyz: Since we're not using `unset_config_file`, let's remove it and add `ENV["CONFIG_FILE_PATH"] = nil` to `delete_config_file`.

rafaelgomesxyz (approved):
Just one small thing, but feel free to merge after changing it.

spec/spec_helper.rb
zzaakiirr: Done, also added `@@tmp_config_file = nil` because this variable should be nil if config file is deleted

PR #220, opened 2024-07-29
Add terraform generate command

zzaakiirr:
### What does this PR do?

This PR introduces new subcommand `terraform` for `cpflow` CLI tool. Currently this subcommand has only 1 command - `generate`

### Important note

This is Draft PR, the main purpose of which is to get opinions on Terraform DSL feature implementation

### DIFF
--- a/lib/core/terraform/config/required_provider.rb
+++ b/lib/core/terraform/config/required_provider.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+module Terraform
+  module Config
+    class RequiredProvider < Base
+      attr_reader :name, :options
+
+      def initialize(name, **options)
+        super()
+
+        @name = name
+        @options = options
+      end
+
+      def to_tf
+        block :terraform do
+          block :required_providers do
+            argument name, options
+          end
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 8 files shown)


lib/core/terraform/config/required_provider.rb
```diff
@@ -0,0 +1,24 @@
...
+
+      def initialize(name, **options)
+        super()
+
+        @name = name
+        @options = options
+      end
+
+      def to_tf
+        block :terraform do
```
zzaakiirr, line 16: @rafaelgomesxyz @borela

Example of DSL usage so that we can describe Terraform config files via `block` and `argument` methods

Do you think that this is the best approach for describing TF config files? Or should we use [Terraform JSON format](https://developer.hashicorp.com/terraform/language/syntax/json)? Or maybe [HEREDOC](https://ruby-doc.org/core-2.5.0/doc/syntax/literals_rdoc.html) will be enough? 🤔 

`Terraform::Config::RequiredProvider#to_tf` produces string (see https://github.com/shakacode/control-plane-flow/pull/220/files#diff-5f77765503204dee0fbde47d7c63260b3d4373530f61680e77f4b13200963b2fR27):
```
terraform  {
  required_providers  {
    cpln = {
        source = "controlplane-com/cpln"
        version = "~> 1.0"
    }
  }
}
```

zzaakiirr:
Closing this PR since terraform features will be implemented incrementally in separate PRs

PR #221, opened 2024-07-30
Introduce terraform subcommand

zzaakiirr:
### What does this PR do?

This PR adds possibility to add subcommands to `cpflow` cli tool and adds `cpflow terraform generate` command which  does nothing (for now)

### Note

- Fixed `CommandHelpers#run_cpflow_command` so that `cpflow help` shows `cpflow <command_name>` i/o `rspec <command_name>`
- `Command::Terraform::Generate` will be implemented in feature PRs

### Screenshots

Screenshot 2024-07-30 at 13 29 22
Screenshot 2024-07-30 at 13 30 06
Screenshot 2024-07-30 at 13 30 38

## Summary by CodeRabbit

- **New Features**
  - Introduced a new `generate` command to automate Terraform configuration file creation in the CLI.
  
- **Bug Fixes**
  - Enhanced command discovery to include files in nested directories, improving the accessibility of commands.

- **Tests**
  - Added a new test case to verify proper handling of subcommands within the CLI.

- **Chores**
  - Improved command execution context management and introduced a utility method for better code organization.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -444,9 +444,9 @@ cpflow run -a $APP_NAME --entrypoint /app/alternative-entrypoint.sh -- rails db:
 cpflow setup-app -a $APP_NAME
 ```
 
-### `generate`
+### `terraform generate`
 
-Generates terraform configuration files based on `controlplane.yml` and `templates/` config
+- Generates terraform configuration files based on `controlplane.yml` and `templates/` config
 
 ```sh
 cpflow terraform generate
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -444,6 +444,14 @@ cpflow run -a $APP_NAME --entrypoint /app/alternative-entrypoint.sh -- rails db:
 cpflow setup-app -a $APP_NAME
 ```
 
+### `generate`
+
+Generates terraform configuration files based on `controlplane.yml` and `templates/` config
+
+```sh
+cpflow terraform generate
+```
+
 ### `version`
 
 - Displays the current version of the CLI
--- a/lib/command/base.rb
+++ b/lib/command/base.rb
@@ -38,16 +38,27 @@ module Command
     WITH_INFO_HEADER = true
     # Which validations to run before the command
     VALIDATIONS = %w[config].freeze
+    SUBCOMMAND = nil
 
     def initialize(config)
       @config = config
     end
 
-    def self.all_commands
-      Dir["#{__dir__}/*.rb"].each_with_object({}) do |file, result|
-        filename = File.basename(file, ".rb")
-        classname = File.read(file).match(/^\s+class (\w+) < Base($| .*$)/)&.captures&.first
-        result[filename.to_sym] = Object.const_get("::Command::#{classname}") if classname
+    def self.all_commands # rubocop:disable Metrics/MethodLength
+      Dir["#{__dir__}/**/*.rb"].each_with_object({}) do |file, result|
+        content = File.read(file)
+
+        classname = content.match(/^\s+class (\w+) < (?:.*Base)(?:$| .*$)/)&.captures&.first
+        next unless classname
+
+        namespaces = content.scan(/^\s+module (\w+)/).flatten
+        full_classname = [*namespaces, classname].join("::").prepend("::")
+
+        command_key = File.basename(file, ".rb")
+        prefix = namespaces[1..1].map(&:downcase).join("_")
+        command_key.prepend(prefix.concat("_")) unless prefix.empty?
+
+        result[command_key.to_sym] = Object.const_get(full_classname)
       end
     end
 
--- a/lib/command/base.rb
+++ b/lib/command/base.rb
@@ -12,6 +12,8 @@ module Command
     VALIDATIONS_WITH_ADDITIONAL_OPTIONS = %w[templates].freeze
     ALL_VALIDATIONS = VALIDATIONS_WITHOUT_ADDITIONAL_OPTIONS + VALIDATIONS_WITH_ADDITIONAL_OPTIONS
 
+    # Used to call the command (`cpflow SUBCOMMAND_NAME NAME`)
+    SUBCOMMAND_NAME = nil
     # Used to call the command (`cpflow NAME`)
     # NAME = ""
     # Displayed when running `cpflow help` or `cpflow help NAME` (defaults to `NAME`)
@@ -38,7 +40,6 @@ module Command
     WITH_INFO_HEADER = true
     # Which validations to run before the command
     VALIDATIONS = %w[config].freeze
-    SUBCOMMAND = nil
 
     def initialize(config)
       @config = config
@@ -55,7 +56,7 @@ module Command
         full_classname = [*namespaces, classname].join("::").prepend("::")
 
         command_key = File.basename(file, ".rb")
-        prefix = namespaces[1..1].map(&:downcase).join("_")
+        prefix = namespaces[1..].map(&:downcase).join("_")
         command_key.prepend(prefix.concat("_")) unless prefix.empty?
 
         result[command_key.to_sym] = Object.const_get(full_classname)
--- a/lib/command/terraform/generate.rb
+++ b/lib/command/terraform/generate.rb
@@ -2,18 +2,13 @@
 
 module Command
   module Terraform
-    class Generate < ::Command::Base
-      SUBCOMMAND = "terraform"
+    class Generate < Base
+      SUBCOMMAND_NAME = "terraform"
       NAME = "generate"
       DESCRIPTION = "Generates terraform configuration files"
       LONG_DESCRIPTION = <<~DESC
-        Generates terraform configuration files based on `controlplane.yml` and `templates/` config
+        - Generates terraform configuration files based on `controlplane.yml` and `templates/` config
       DESC
-      EXAMPLES = <<~EX
-        ```sh
-        cpflow terraform generate
-        ```
-      EX
       WITH_INFO_HEADER = false
       VALIDATIONS = [].freeze
 
--- a/lib/command/terraform/generate.rb
+++ b/lib/command/terraform/generate.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module Command
+  module Terraform
+    class Generate < ::Command::Base
+      SUBCOMMAND = "terraform"
+      NAME = "generate"
+      DESCRIPTION = "Generates terraform configuration files"
+      LONG_DESCRIPTION = <<~DESC
+        Generates terraform configuration files based on `controlplane.yml` and `templates/` config
+      DESC
+      EXAMPLES = <<~EX
+        ```sh
+        cpflow terraform generate
+        ```
+      EX
+      WITH_INFO_HEADER = false
+      VALIDATIONS = [].freeze
+
+      def call
+        # TODO: Implement
+      end
+    end
+  end
+end
--- a/lib/cpflow.rb
+++ b/lib/cpflow.rb
@@ -113,6 +113,10 @@ module Cpflow
     def self.fix_help_option
       help_mappings = Thor::HELP_MAPPINGS + ["help"]
       matches = help_mappings & ARGV
+
+      # Help option works correctly for subcommands
+      return if matches && (ARGV & subcommand_names).any?
+
       matches.each do |match|
         ARGV.delete(match)
         ARGV.unshift(match)
@@ -149,6 +153,10 @@ module Cpflow
       ::Command::Base.all_commands.merge(deprecated_commands)
     end
 
+    def self.subcommand_names
+      ::Command::Base.all_commands.values.map { |command| command::SUBCOMMAND }.compact
+    end
+
     def self.process_option_params(params)
       # Ensures that if no value is provided for a non-boolean option (e.g., `cpflow command --option`),
       # it defaults to an empty string instead of the option name (which is the default Thor behavior)
@@ -156,6 +164,17 @@ module Cpflow
 
       params
     end
+
+    def self.klass_for(subcommand)
+      klass_name = subcommand.to_s.split("-").collect(&:capitalize).join
+      return Cpflow.const_get(klass_name) if Cpflow.const_defined?(klass_name)
+
+      Cpflow.const_set(klass_name, Class.new(BaseSubCommand)).tap do |subcommand_klass|
+        desc subcommand, "#{subcommand.capitalize} commands"
+        subcommand subcommand, subcommand_klass
+      end
+    end
+    private_class_method :klass_for
 
     @commands_with_required_options = []
     @commands_with_extra_options = []
@@ -181,6 +200,7 @@ module Cpflow
       hide = command_class::HIDE || deprecated
       with_info_header = command_class::WITH_INFO_HEADER
       validations = command_class::VALIDATIONS
+      subcommand = command_class::SUBCOMMAND
 
       long_description += "\n#{examples}" if examples.length.positive?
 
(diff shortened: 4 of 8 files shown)


zzaakiirr:
@coderabbitai review

lib/command/terraform/generate.rb
```diff
@@ -0,0 +1,25 @@
...
+        ```sh
+        cpflow terraform generate
+        ```
+      EX
+      WITH_INFO_HEADER = false
+      VALIDATIONS = [].freeze
+
+      def call
+        # TODO: Implement
+      end
```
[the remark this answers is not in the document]
zzaakiirr: > Do you want me to help with the implementation of this method or open a GitHub issue to track this task?

No, since this method will be implemented in next PR

borela (approved):

docs/commands.md
```diff
@@ -444,6 +444,14 @@ cpflow run -a $APP_NAME --entrypoint /app/alternative-entrypoint.sh -- rails db:
 cpflow setup-app -a $APP_NAME
 ```
 
+### `generate`
```
rafaelgomesxyz, line 447: We need to fix the script that generates this file to take into account subcommands.

lib/command/base.rb
```diff
@@ -38,16 +38,27 @@ class Base # rubocop:disable Metrics/ClassLength
...
     def initialize(config)
       @config = config
     end
 
-    def self.all_commands
-      Dir["#{__dir__}/*.rb"].each_with_object({}) do |file, result|
-        filename = File.basename(file, ".rb")
-        classname = File.read(file).match(/^\s+class (\w+) < Base($| .*$)/)&.captures&.first
-        result[filename.to_sym] = Object.const_get("::Command::#{classname}") if classname
+    def self.all_commands # rubocop:disable Metrics/MethodLength
```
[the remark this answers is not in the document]
rafaelgomesxyz: No, we're not defining the commands in each command file, we're just configuring them. Thor doesn't handle it like that. So we need to load all of the configurations and manually define the Thor commands, which is done in: https://github.com/shakacode/control-plane-flow/blob/66e532bf69e2d2136b15b3608eb9c5007120a0fb/lib/cpflow.rb#L168

lib/command/terraform/generate.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Command
+  module Terraform
+    class Generate < ::Command::Base
```
rafaelgomesxyz, line 5: Can we not do just `< Base`?

lib/command/terraform/generate.rb
```diff
@@ -0,0 +1,25 @@
...
+      SUBCOMMAND = "terraform"
+      NAME = "generate"
+      DESCRIPTION = "Generates terraform configuration files"
+      LONG_DESCRIPTION = <<~DESC
+        Generates terraform configuration files based on `controlplane.yml` and `templates/` config
```
rafaelgomesxyz, line 10: For `LONG_DESCRIPTION` we generally use bullet points.

```suggestion
        - Generates terraform configuration files based on `controlplane.yml` and `templates/` config
```

lib/cpflow.rb
```diff
@@ -113,6 +113,10 @@ def self.check_cpflow_version # rubocop:disable Metrics/MethodLength
     def self.fix_help_option
       help_mappings = Thor::HELP_MAPPINGS + ["help"]
       matches = help_mappings & ARGV
+
+      # Help option works correctly for subcommands
+      return if matches && (ARGV & subcommand_names).any?
```
rafaelgomesxyz, line 118: I don't think we're supposed to return here. Did you test if `cpflow SUBCOMMAND COMMAND --help` works?

lib/cpflow.rb
```diff
@@ -149,6 +153,10 @@ def self.all_base_commands
       ::Command::Base.all_commands.merge(deprecated_commands)
     end
 
+    def self.subcommand_names
+      ::Command::Base.all_commands.values.map { |command| command::SUBCOMMAND }.compact
```
rafaelgomesxyz, line 157: We could also just read the names of the subfolders in the `command` folder.

lib/cpflow.rb
```diff
@@ -181,28 +200,33 @@ def self.process_option_params(params)
...
+        desc(usage, description, hide: hide)
+        long_desc(long_description)
+
+        command_options.each do |option|
+          params = process_option_params(option[:params])
+          method_option(option[:name], **params)
+        end
+      end
+
+      klass.define_method(name_for_method) do |*provided_args| # rubocop:disable Metrics/BlockLength, Metrics/MethodLength
```
rafaelgomesxyz, line 229: This is because we need to define it in the subcommand class, correct? There might be some other things we do here that we need to double-check if we also need to do in the subcommand class for them to work.

spec/support/command_helpers.rb
```diff
@@ -157,6 +157,9 @@ def create_app_if_not_exists(app, deploy: false, image_before_deploy_count: 0, i
   end
 
   def run_cpflow_command(*args, raise_errors: false) # rubocop:disable Metrics/MethodLength
+    program_name_before = $PROGRAM_NAME
+    $PROGRAM_NAME = package_name
```
rafaelgomesxyz, line 161: What's the reason for this? And where does `$PROGRAM_NAME` come from? Thor?

rafaelgomesxyz (changes requested):
Just a partial review so far, but I almost wonder if it wouldn't be easier to just use `:` like we do for `ps:start`, `ps:stop`, etc. Like `terraform:generate`. Are subcommands preferrable?

lib/command/base.rb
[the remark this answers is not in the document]
zzaakiirr: It seems that we need this method to be able to use classes i/o methods. In order to register a new coomand we need to do something like this:
```
class SomeClass < Thor
  desc "usage", "desc"
  long_desc "long desc"
  def command_name
  end
end
```
But we are creating separate classes for each command: 
```
module Command
  class SomeCommand < Base
    NAME = "command_name"
    USAGE = "usage"
    DESCRIPTION = "desc"
    LONG_DESCRIPTION = "long desc"

    def call
      ...
    end
  end
end
```

Which later transform to methods:
```
class SomeClass < Thor
  all_commands.each do |command_key, command_class|
    name = command_class::NAME
    usage = command_class::USAGE
    description = command_class::DESCRIPTION
    long_description = command_class::LONG_DESCRIPTION

    desc(usage, description, hide: hide)
    long_desc(long_description)
    define_method(name) do
      command_class.new.call
    end
  end
end
```

lib/cpflow.rb
zzaakiirr: Yes, it works because of `BaseSubCommand` class:
```
./cpflow terraform generate --help
```
Results in:
```
Usage:
  cpflow terraform generate

Description:
  Generates terraform configuration files based on `controlplane.yml` and `templates/` config

  cpflow terraform generate
```

lib/cpflow.rb
zzaakiirr: > This is because we need to define it in the subcommand class, correct? 

Yes

> There might be some other things we do here that we need to double-check if we also need to do in the subcommand class for them to work.

Good point, I'll check if I missed something

spec/support/command_helpers.rb
zzaakiirr: It comes from Thor - https://github.com/rails/thor/blob/a43d92fad7ebd77d359b7b96eb3db8a73ef9057c/lib/thor/base.rb#L772

Without this tweak we get this output (`rspec` i/o `cpflow`):
```
cpflow commands:
  rspec apply-template TEMPLATE [TEMPLATE] ... [TEMPLATE] -a, --app=APP_NAME              # Applies application-specific configs from templates
  rspec build-image -a, --app=APP_NAME                                                    # Builds and pushes the image to Control Plane
  rspec cleanup-images -a, --app=APP_NAME                                                 # Deletes all images for an app that either exceed the max quanti...
  rspec cleanup-stale-apps -a, --app=APP_NAME                                             # Deletes the whole app (GVC with all workloads, all volumesets a...
  rspec config                                                                            # Displays config for each app or a specific app
  rspec copy-image-from-upstream -a, --app=APP_NAME -t, --upstream-token=UPSTREAM_TOKEN   # Copies an image (by default the latest) from a source org to th...
  rspec delete -a, --app=APP_NAME                                                         # Deletes the whole app (GVC with all workloads, all volumesets a...
  rspec deploy-image -a, --app=APP_NAME                                                   # Deploys the latest image to app workloads, and runs a release s...
  rspec doctor                                                                            # Runs validations
  rspec env -a, --app=APP_NAME                                                            # Displays app-specific environment variables
  rspec exists -a, --app=APP_NAME                                                         # Shell-checks if an application (GVC) exists, useful in scripts
  rspec generate                                                                          # Creates base Control Plane config and template files
  rspec help [COMMAND]                                                                    # Describe available commands or one specific command
  rspec info                                                                              # Displays the diff between defined/available apps/workloads (app...
  rspec latest-image -a, --app=APP_NAME                                                   # Displays the latest image name
  rspec logs -a, --app=APP_NAME                                                           # Light wrapper to display tailed raw logs for app/workload syntax
  rspec maintenance -a, --app=APP_NAME                                                    # Checks if maintenance mode is on or off for an app
  rspec maintenance:off -a, --app=APP_NAME                                                # Disables maintenance mode for an app
  rspec maintenance:on -a, --app=APP_NAME                                                 # Enables maintenance mode for an app
  rspec maintenance:set-page PAGE_URL -a, --app=APP_NAME                                  # Sets the page for maintenance mode
  rspec open -a, --app=APP_NAME                                                           # Opens the app endpoint URL in the default browser
  rspec open-console -a, --app=APP_NAME                                                   # Opens the app console on Control Plane in the default browser
  rspec promote-app-from-upstream -a, --app=APP_NAME -t, --upstream-token=UPSTREAM_TOKEN  # Copies the latest image from upstream, runs a release script (o...
  rspec ps -a, --app=APP_NAME                                                             # Shows running replicas in app
  rspec ps:restart -a, --app=APP_NAME                                                     # Forces redeploy of workloads in app
  rspec ps:start -a, --app=APP_NAME                                                       # Starts workloads in app
  rspec ps:stop -a, --app=APP_NAME                                                        # Stops workloads in app
  rspec ps:wait -a, --app=APP_NAME                                                        # Waits for workloads in app to be ready after re-deployment
  rspec run COMMAND -a, --app=APP_NAME                                                    # Runs one-off interactive or non-interactive replicas (analog of...
  rspec setup-app -a, --app=APP_NAME                                                      # Creates an app and all its workloads
  rspec terraform                                                                         # Terraform commands
  rspec version                                                                           # Displays the current version of the CLI

Options:
  -o, [--org=ORG_NAME]             # Organization name
  -d, [--verbose], [--no-verbose]  # Shows detailed logs
      [--trace], [--no-trace]      # Shows trace of API calls. WARNING: may contain sensitive data
```

zzaakiirr:
> Just a partial review so far, but I almost wonder if it wouldn't be easier to just use : like we do for ps:start, ps:stop, etc. Like terraform:generate. Are subcommands preferrable?

I missed that 🥲

I still prefer subcommands because user can get all `terraform` commands by invoking `cpflow terraform --help`. The problem with `:` approach is that user needs to invoke `cpflow help` for each command in this namespace

Another issue with this approach is class names which must include subcommand namespace in class names (`class Ps<CommandName>` for example), declaring separate module is better (IMHO)

I won't push on subcommand approach @rafaelgomesxyz @borela WDYT?

lib/command/base.rb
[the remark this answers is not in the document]
borela: @rafaelgomesxyz Why do we need to do it that way instead of reopening the main class and load all files on the [entrypoint](https://github.com/shakacode/control-plane-flow/blob/main/lib/cpflow.rb)?

This way each command would be declared normally like Thor expect us to do.

borela:
@zzaakiirr I am in favor of your design as it is more common in UNIX for CLIs that support subcommands.

@justin808 @rafaelgomesxyz @dzirtusss @zzaakiirr I would prefer to change the loading logic to just reopen the CLI class for each command and declare them like Thor docs says, e.g.:

```
# command_1.rb
class CLI < Thor
  desc "usage", "desc"
  long_desc "long desc"
  def command_1
  end
end

# command_2.rb
class CLI < Thor
  desc "usage", "desc"
  long_desc "long desc"
  def command_2
  end
end

...
```

By reopening the class this way, you can still share state, methods etc... and still follow Thor's design.

rafaelgomesxyz:
@borela We have a bunch of custom logic that benefits from the current loading logic, as you can see in https://github.com/shakacode/control-plane-flow/blob/main/lib/cpflow.rb#L168.

What are the benefits of switching to your proposed logic?

Thor doesn't really has a design for defining commands in separate files, so I'm not sure what you mean by that.

I think we should keep it the way it is.

borela:
@rafaelgomesxyz, it looks like you are re-implementing Thor's features; Which features that custom logic implement that Thor already doesn't support? Validations? That could be extracted to a method that accepts a value and pattern to validate against.

rafaelgomesxyz:
@borela We're not though, we just have a layer on top of Thor to make things easier. It's not just validations, but also handling of deprecated commands and special args / options cases, as well as setting some fallback values here and there.

Regardless, if we were to reopen the class in every command file, we'd have to repeat all of this logic for each file. Sure, we could move some logic to common helpers, but we'd still have repetition, which is why this way seems better to me. I don't see any benefits in switching to the way you proposed.

borela:
The code for each command would be linear and less magical where you see the command being defined and all options it has (could use some methods to prevent repetition here like you said but just call Thor's option method making sure self is the CLI class). 

A little bit of repetition is fine if the end result is simpler to read.

borela:
@rafaelgomesxyz The code for each command would be linear and less magical. A little bit of repetition is fine if the end result is simpler to read.

In an example scenario, command x is having a validation issue. I will go to the command x file and see how validation is being performed (either directly or by calling an accessory function) inside the command's method.

That does not happen with the current design, it is not immediately obvious where simple stuff like validations are happening or that those options are being validated. In contrast, other projects using thor just do simple ||= for fallback values and validate the option on the command itself, which again can be extracted to a method and called by the command.

rafaelgomesxyz:
Let's see what Sergey and Justin think, but I still think we should keep it as is.

I think once you understand the constants defined at the top of the command classes and how options are defined in `Base`, it becomes pretty straightforward to figure out how something is being validated.

Another thing I like about the current approach is that we have a somewhat centralized place for all Thor stuff, which allows us to have a better separation of concerns, and detaches our app from Thor. It would be a lot easier to adapt the code if we wanted to replace Thor with something else - I know Justin was considering moving to an `npm` package so we have both the CPLN CLI and our CLI in the same place, and although that wouldn't help too much in this case, because we'd still need to convert from Ruby to JavaScript, I think it would definitely make things easier.

We also have some configs set on the class level based on the commands, which benefits from the current approach:

https://github.com/shakacode/control-plane-flow/blob/82ff517297fa79e066192d9d298f32c8820b77d2/lib/cpflow.rb#L246-L248

These cannot be set individually, so we'd still need a way to go through all command classes looking for matches.

lib/command/base.rb
```diff
@@ -38,16 +38,27 @@ class Base # rubocop:disable Metrics/ClassLength
...
+        content = File.read(file)
+
+        classname = content.match(/^\s+class (\w+) < (?:.*Base)(?:$| .*$)/)&.captures&.first
+        next unless classname
+
+        namespaces = content.scan(/^\s+module (\w+)/).flatten
+        full_classname = [*namespaces, classname].join("::").prepend("::")
+
+        command_key = File.basename(file, ".rb")
+        prefix = namespaces[1..1].map(&:downcase).join("_")
```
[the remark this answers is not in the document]
zzaakiirr: You are right, I meant  `namespaces[1..]`

docs/commands.md
zzaakiirr: Fixed the script

lib/command/terraform/generate.rb
zzaakiirr: Yes, we can - done

lib/cpflow.rb
zzaakiirr: Created class method `subcommand?`

lib/cpflow.rb
zzaakiirr: Checked again - didn't find anything new

spec/support/command_helpers.rb
zzaakiirr: Reverted this change

spec/cpflow_spec.rb
```diff
@@ -19,4 +19,19 @@
...
     end
   end
+
+  it "handles subcommands correctly" do
+    result = run_cpflow_command("help")
+
+    expect(result[:status]).to eq(0)
+
+    Cpflow::Cli.subcommand_names.each do |subcommand|
+      expect(result[:stdout]).to include("#{package_name} #{subcommand}")
```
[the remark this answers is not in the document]
zzaakiirr: Replaced `package_name` with `basename`

lib/command/terraform/generate.rb
```diff
@@ -0,0 +1,25 @@
...
+      DESC
+      EXAMPLES = <<~EX
+        ```sh
+        cpflow terraform generate
```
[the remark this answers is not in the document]
zzaakiirr: Removed `EXAMPLES` constant

borela:
@rafaelgomesxyz Looking at the source, just [build image has ACCEPTS_EXTRA_OPTIONS](https://github.com/search?q=repo%3Ashakacode%2Fcontrol-plane-flow%20ACCEPTS_EXTRA_OPTIONS&type=code) set to true, and it looks like it only does this because it forwards them to docker.

Why not declare the docker options supported? The error message will be better and when somebody does `cpflow build --help` they will see them too.

borela:
@rafaelgomesxyz if zakir pr is following the current pattern and working we can merge this pr and I can move my suggestions to an issue.

lib/command/base.rb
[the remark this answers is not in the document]
rafaelgomesxyz: It's still `namespaces[1..1]` by the way.

rafaelgomesxyz (approved):
@zzaakiirr Just one thing that you said you changed, but probably forgot to. Feel free to merge after changing it.

PR #229, opened 2024-08-30
Maintenance mode: check if domain is updated via API

zzaakiirr:
### What does this PR do?

This PR replaces `sleep` with actual API request to check if `workloadLink` is changed for domain after update

### Tests

Specs for maintenance command are passing - https://github.com/shakacode/control-plane-flow/actions/runs/10626502289/job/29458117242

## Summary by CodeRabbit

- **New Features**
	- Enhanced the maintenance workload switching process with a dynamic retry mechanism for improved responsiveness.
	- Introduced error handling to ensure domain availability during workload switches.

- **Bug Fixes**
	- Replaced static wait time with a more efficient verification approach to confirm the workload state.

### DIFF
--- a/lib/core/maintenance_mode.rb
+++ b/lib/core/maintenance_mode.rb
@@ -84,9 +84,13 @@ class MaintenanceMode
   def switch_domain_workload(to:)
     step("Switching workload for domain '#{domain_data['name']}' to '#{to}'") do
       cp.set_domain_workload(domain_data, to)
+    end
 
-      # Give it a bit of time for the domain to update
-      Kernel.sleep(30)
+    step("Waiting for changes to take effect", retry_on_failure: true, wait: 10, max_retry_count: 3) do
+      refetched_domain_data = cp.fetch_domain(domain_data["name"])
+      raise "Can't find domain" if refetched_domain_data.nil?
+
+      cp.domain_workload_matches?(refetched_domain_data, to)
     end
 
     progress.puts


zzaakiirr:
@coderabbitai review

lib/core/maintenance_mode.rb
```diff
@@ -84,9 +84,13 @@ def start_or_stop_maintenance_workload(action)
...
       cp.set_domain_workload(domain_data, to)
+    end
+
+    step("Waiting for changes to take effect", retry_on_failure: true, wait: 10, max_retry_count: 3) do
+      refetched_domain_data = cp.fetch_domain(domain_data["name"])
+      raise "Can't find domain" if refetched_domain_data.nil?
 
-      # Give it a bit of time for the domain to update
-      Kernel.sleep(30)
+      domain_workload_matches?(refetched_domain_data, to)
```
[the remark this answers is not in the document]
zzaakiirr: `config` has no such options and method `step` has default values for retry parameters, here we just provide our own `max_retry_count` & `wait`

lib/core/maintenance_mode.rb
```diff
@@ -84,9 +84,13 @@ def start_or_stop_maintenance_workload(action)
   def switch_domain_workload(to:)
     step("Switching workload for domain '#{domain_data['name']}' to '#{to}'") do
       cp.set_domain_workload(domain_data, to)
+    end
+
+    step("Waiting for changes to take effect", retry_on_failure: true, wait: 10, max_retry_count: 3) do
```
justin808, line 89: @borela should we be using https://github.com/shakacode/uber_task?

@zzaakiirr is this ready to merge?

lib/core/maintenance_mode.rb
```diff
@@ -84,9 +84,13 @@ def start_or_stop_maintenance_workload(action)
...
+      refetched_domain_data = cp.fetch_domain(domain_data["name"])
+      raise "Can't find domain" if refetched_domain_data.nil?
 
-      # Give it a bit of time for the domain to update
-      Kernel.sleep(30)
+      cp.domain_workload_matches?(refetched_domain_data, to)
```
zzaakiirr, line 93: @rafaelgomesxyz Is this check enough to be sure that domain was updated?

lib/core/maintenance_mode.rb
zzaakiirr: @justin808 I left a question for @rafaelgomesxyz, I want to get his opinion on approach I took in this PR for checking if domain was updated.

If this is the right way and Rafael approves this PR then we can merge this

lib/core/maintenance_mode.rb
zzaakiirr: > So we'd need a different method for checking it 

Do you have any ideas how can we check this? Maybe controlplane displays some message while workload change takes effect?

rafaelgomesxyz: @zzaakiirr A way I can think of is to make a real request to the domain and see if it's displaying the regular page or the maintenance page.

I'm not sure if there's a better way to do it.

PR #243, opened 2024-11-12
Terraform feature fixes

zzaakiirr:
### What does this PR do?

This PR fixes errors that were discovered while generating terraform configurations for **hichee** app


lib/core/terraform_config/dsl.rb
```diff
@@ -30,6 +30,9 @@ def argument(name, value, optional: false, raw: false)
           "#{tf_value(value)}\n"
         end
 
+      # remove quotes from expression values
+      content = content.gsub(/("#{EXPRESSION_PATTERN}.*")/) { ::Regexp.last_match(1)[1...-1] }
```
zzaakiirr, line 34: `"['cpln_secret.some-secret.name']"` -> `"[cpln_secret.some-secret.name]"`

lib/core/terraform_config/generator.rb
```diff
@@ -107,7 +108,7 @@ def volumeset_config_params
     def workload_config_params
       template
         .slice(:name, :description, :tags)
-        .merge(gvc: gvc, identity: workload_identity)
```
zzaakiirr, line 110: using `identity_link` i/o `identity` since there is a case when `identity` is referenced via link in workload config and `.controlplane/templates/` does not have this identity config

zzaakiirr:
@coderabbitai review

spec/core/terraform_config/gvc_spec.rb
```diff
@@ -34,7 +34,7 @@
               }
               domain = "app.example.com"
               locations = ["aws-us-east-1", "aws-us-east-2"]
-              pull_secrets = ["cpln_secret.docker.name"]
+              pull_secrets = [cpln_secret.docker.name]
```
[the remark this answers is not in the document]
zzaakiirr: It is ok - string literal is transformed to resource reference in `TerraformConfig::Dsl#argument`

spec/core/terraform_config/workload_spec.rb
```diff
@@ -9,7 +9,7 @@
       description: "main workload description",
       tags: { tag1: "tag1_value", tag2: "tag2_value" },
       gvc: "cpln_gvc.app-name.name",
-      identity: "cpln_identity.identity-name",
+      identity_link: "//gvc/gvc-name/identity/app-identity",
```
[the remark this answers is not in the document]
zzaakiirr: We need to change `dentity` to `identity_link` only for `workload` template

borela (approved):

.gitignore
```diff
@@ -18,5 +18,5 @@
 /spec/dummy/.controlplane/controlplane*-tmp-*.yml
 
 # Generated configs
-terraform/
-.controlplane/
+/terraform/
```
zzaakiirr, line 21: We need to ignore `.controlplane` folder on root level, otherwise folder `spec/dummy/.controlplane` is also ignored

lib/core/terraform_config/generator.rb
rafaelgomesxyz: But doesn't this mean that we're excluding one case to support the other? Can we not support both cases? Or do we not need to? Or did I get it wrong?

Other than that, this PR looks good to me.

zzaakiirr: I the use `identity` as reference then identity should be defined in `.controlplane/templates`. Having `identity_link` is more general solution because:
1. We don't require identity definition in `templates/` folder
2. Workload template (in YAML format) defines usage of `identity` via `identityLink` so it's easier to just use it i/o creating reference to identity from this link

rafaelgomesxyz: Got it, all good then.

rafaelgomesxyz (approved):
LGTM

PR #245, opened 2024-11-25
Delete workloads before volumesets on GVC deletion

zzaakiirr:
### What does this PR do?

This PR fixes issue where GVC deletion fails because one of the workloads has volumeset in-use - added deletion of workloads before deletion of volumesets

### Steps to reproduce issue

1. Create GVC
2. Create volumeset in GVC
3. Create workload which uses volumeset (created in step 2)
4. Run `cpflow delete -a <GVC_NAME>`

### Expected behaviour

All volumesets, workloads, identities and images and GVC are deleted

### Actual behaviour

Error `This volume set is in use by a workload and cannot be deleted` appears

### Notes

The idea to delete workloads before volumesets is taken from ControlPlane UI - when we force delete GVC multiple API requests are fired:
1. Gathering and deletion of workloads
2. Gathering and deletion of volumesets
3. Gathering and deletion of identites
4. Gathering and deletion of images
5. Deletion of the GVC


zzaakiirr:
@coderabbitai review

lib/command/delete.rb
```diff
@@ -51,17 +51,37 @@ def delete_single_workload(workload)
...
     end
 
+    def check_workloads
+      @workloads = cp.fetch_workloads["items"]
+      return progress.puts("No workloads to delete from app '#{config.app}'.") unless @workloads.any?
+
+      message = "The following workloads will be deleted along with the app '#{config.app}':"
+      workloads_list = @workloads.map { |workload| "- #{workload['name']}" }.join("\n")
+      progress.puts("#{Shell.color(message, :red)}\n#{workloads_list}\n\n")
+    end
```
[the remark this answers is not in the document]
zzaakiirr: `cp.fetch_workloads` returns `[]` if GVC has no workloads

zzaakiirr:
@dzirtusss 

> Can we look for force option in api/cli? This will guarantee deletion.

There is no `force` option in API docs:
image

But there is `--force` option in `cpln` cli:
Screenshot 2024-11-26 at 13 02 30

Current implementation of `cpflow delete` logic relies on API requests - I think it's possible to leave gathering info about items that will be deleted via API and replace API DELETE requests with `cpln gvc delete <ref> --force` command if whole app needs to deleted.

Should we take advantage of this command i/o multiple API DELETE requests? @dzirtusss @rafaelgomesxyz @dzirtusss  WDYT?

rafaelgomesxyz:
@zzaakiirr I'm confused. I had already fixed this issue:

image

It checks if a volumeset is attached to a workload and deletes it first.

Why is it not working?

rafaelgomesxyz:
@zzaakiirr @dzirtusss By the way, I prefer if we rely on the API more than the CLI, because we can run into versioning issues with the CLI, while that's not the case for the API.

zzaakiirr:
> @zzaakiirr I'm confused. I had already fixed this issue:
> 
> image
> 
> It checks if a volumeset is attached to a workload and deletes it first.
> 
> Why is it not working?

Sorry, I missed that. Maybe API response changed? I don't seee `status -> usedByWorkload` in response, only `status -> workloadLinks`:
image

If so, we can fix only this part:
image

OR we can leave current implementation and remove workload deletion logic from `#delete_volumesets` - this way user can see which workloads will be removed but execution will take longer since we always will be removing all workloads.

@rafaelgomesxyz WDYT?

lib/command/delete.rb
```diff
@@ -106,9 +106,9 @@ def delete_workload(workload)
     def delete_volumesets
       @volumesets.each do |volumeset|
         step("Deleting volumeset '#{volumeset['name']}' from app '#{config.app}'") do
-          # If the volumeset is attached to a workload, we need to delete the workload first
-          workload = volumeset.dig("status", "usedByWorkload")&.split("/")&.last
-          cp.delete_workload(workload) if workload
+          # If the volumeset is attached to a workloads, we need to delete the workloads first
```
rafaelgomesxyz, line 109: Minor typo:

```suggestion
          # If the volumeset is attached to workloads, we need to delete the workloads first
```

CHANGELOG.md
```diff
@@ -20,6 +20,8 @@ _Please add entries here for your pull requests that have not yet been released.
 
 - Fixed potential infinite loop that could occur for a command if one of the execution steps fails and gets stuck. [PR 217](https://github.com/shakacode/control-plane-flow/pull/217) by [Zakir Dzhamaliddinov](https://github.com/zzaakiirr).
 
+- Fixed issue where GVC cannot be deleted because one of the workloads has volumeset in-use. [PR 245](https://github.com/shakacode/control-plane-flow/pull/245) by [Zakir Dzhamaliddinov](https://github.com/zzaakiirr).
```
rafaelgomesxyz, line 23: Minor suggestion: we generally use "app" instead of "GVC" in the docs.

```suggestion
- Fixed issue where app cannot be deleted because one of the workloads has a volumeset in-use. [PR 245](https://github.com/shakacode/control-plane-flow/pull/245) by [Zakir Dzhamaliddinov](https://github.com/zzaakiirr).
```

rafaelgomesxyz (approved):
LGTM. Left just a couple minor suggestions, if you could do those before merging.

PR #256, opened 2025-05-05
Add node_no_warnings environment variable

Judahmeek:
While figuring out how to test this locally, I also encountered [a really old error with the debug gem that should have been resolved a long time ago](https://github.com/ruby/debug/issues/275) (I'm using Ruby 2.7.3).

### DIFF
--- a/lib/cpflow.rb
+++ b/lib/cpflow.rb
@@ -51,6 +51,7 @@ module Cpflow
 
     def self.start(*args)
       ENV["CPLN_SKIP_UPDATE_CHECK"] = "true"
+      # ENV["NODE_NO_WARNINGS"] = "1"
 
       check_cpln_version
       check_cpflow_version


lib/cpflow.rb
```diff
@@ -51,6 +51,7 @@ class Cli < Thor # rubocop:disable Metrics/ClassLength
 
     def self.start(*args)
       ENV["CPLN_SKIP_UPDATE_CHECK"] = "true"
+      # ENV["NODE_NO_WARNINGS"] = "1"
```
rafaelgomesxyz, line 54: It's commented out.

rafaelgomesxyz (changes requested):

lib/cpflow.rb
Judahmeek: Fixed. That was a little embarrassing.

rafaelgomesxyz (approved):

PR #404, opened 2026-07-10
Fix cpflow loading without a UTF-8 locale

### LINKED ISSUE #372: cpflow crashes at startup on systems without a UTF-8 locale (File.read without encoding in Command::Base.all_commands)
## Summary

Every `cpflow` invocation crashes at require time with `invalid byte sequence in US-ASCII (ArgumentError)` when the process has no UTF-8 locale (`LANG`/`LC_ALL` unset, or `C`/`POSIX`). This hits minimal Docker images, cron jobs, some CI runners, and sandboxed shells — the CLI is completely unusable there, including `cpflow version` and `cpflow --help`.

## Root cause

[`Command::Base.all_commands`](https://github.com/shakacode/control-plane-flow/blob/7c96df930980982160da75d22716202a82a94838/lib/command/base.rb#L51-L54) reads every file under `lib/command/**` with `File.read(file)` (no explicit encoding) and runs a regex over the content. `File.read` without an encoding uses `Encoding.default_external`, which Ruby derives from the locale — with no UTF-8 locale it becomes `US-ASCII`. Two command files contain non-ASCII characters, so `String#match` raises:

- `lib/command/cleanup_stale_apps.rb`
- `lib/command/delete.rb`

The crash happens at **load time**, because `all_commands` is invoked while `Cpflow::Cli` is being defined ([lib/cpflow.rb#L214](https://github.com/shakacode/control-plane-flow/blob/7c96df930980982160da75d22716202a82a94838/lib/cpflow.rb#L214) via [#L250](https://github.com/shakacode/control-plane-flow/blob/7c96df930980982160da75d22716202a82a94838/lib/cpflow.rb#L250)). Every command is affected.

## Reproduction (verified on 7c96df9, Ruby 3.4.6, macOS)

```sh
LC_ALL= LANG= bundle exec ruby -e 'require_relative "lib/cpflow"'
```

```
lib/command/base.rb:54:in 'Regexp#match': invalid byte sequence in US-ASCII (ArgumentError)
  from lib/command/base.rb:54:in 'block in Command::Base.all_commands'
  ...
  from lib/cpflow.rb:250:in '<class:Cli>'
```

Also reproduces via `LC_ALL= LANG= ./script/check_command_docs`. CI never catches this because GitHub-hosted runners export a UTF-8 locale.

## Proposed fix

1. Read with explicit encoding in `all_commands`: `File.read(file, encoding: "UTF-8")`.
2. Audit the other `File.read` call sites in `lib/` that parse file or template content (`grep -rn "File.read" lib`) for the same latent issue.
3. Add a regression spec: assert `Command::Base.all_commands` returns the full command set when `Encoding.default_external` is `US-ASCII` (run in a subprocess with `ruby -E US-ASCII:US-ASCII`, or set/restore `Encoding.default_external` around the call with `$VERBOSE` silenced).

## Acceptance criteria

- [ ] `LC_ALL= LANG= bundle exec ruby -e 'require_relative "lib/cpflow"'` exits 0
- [ ] `cpflow version` works in a container with no locales installed
- [ ] Regression spec covering the non-UTF-8 `default_external` case
- [ ] `CHANGELOG.md` entry under Unreleased → Fixed (user-visible bug)

## Verification

`bundle exec rubocop` plus the repro command above. CI fast suite runs on the PR automatically.

## Priority / size / dependencies

**P1** — user-facing startup crash with a one-line fix. Size: **S**. Dependencies: **none** (good first candidate to land).

---
_Origin: automated codebase health review, 2026-07-02, against 7c96df9. Reproduced locally — not theoretical._

justin808:
## Summary

- Read command files as UTF-8 so command discovery does not depend on the process locale.
- Add a regression that requires cpflow under the C locale, where Ruby uses US-ASCII as its external encoding.
- Document the user-visible startup fix.

Fixes #372.

## Attribution

Independently recreated from the narrow proposal in #398 by @SAY-5. The fork branch was not checked out or executed; the maintainer branch contains a reviewed implementation and a stronger public-interface regression test.

## Validation

- PASS: bundle exec rspec spec/cpflow_spec.rb
- PASS: LC_ALL=C LANG=C bundle exec ruby -I lib -e require cpflow
- PASS: .agents/bin/lint
- BLOCKED by environment: .agents/bin/validate requires a real CPLN_ORG for live Control Plane specs. Its failures are the documented missing-org errors and are unrelated to this locale-only change.

## Review

- Manual diff review: no actionable findings.
- Local Codex review was unavailable because the installed CLI does not support its configured model. Claude Code review was unavailable because this environment is not logged in.

## Codex Decision Log

- Non-blocking: adopt the external proposal on a maintainer-owned branch.
  - Decision: recreate and validate the minimal fix here rather than merge or modify the fork.
  - Why: preserves the trust boundary while retaining contributor attribution.
  - Review later: None.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a startup crash when running on systems using a non-UTF-8 locale.
  * Command files are now loaded consistently using UTF-8 encoding.

* **Tests**
  * Added coverage to verify successful startup under US-ASCII locale settings.

* **Documentation**
  * Documented the locale-related startup fix in the unreleased changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ In addition to the standard keepachangelog.com categories, this project uses a l
 - **Added generic telemetry documentation for deploying an OpenTelemetry Collector with Control Plane Flow**, including collector workload templates, application instrumentation, telemetry pipelines, review-app isolation, and troubleshooting guidance. [PR 369](https://github.com/shakacode/control-plane-flow/pull/369) by [Justin Gordon](https://github.com/justin808).
 - **Added a Rails-focused Grafana and OpenTelemetry guide for building Control Plane dashboards from generated span and log metrics**, including collector workload guidance, spanmetrics setup, rollout order, alerting, and validation checklists. [PR 352](https://github.com/shakacode/control-plane-flow/pull/352) by [Justin Gordon](https://github.com/justin808).
 
+### Fixed
+
+- **Fixed `cpflow` crashing at load time with `invalid byte sequence in US-ASCII (ArgumentError)` on systems without a UTF-8 locale.** `Command::Base.all_commands` now reads command files with an explicit UTF-8 encoding instead of relying on `Encoding.default_external`. Fixes [issue 372](https://github.com/shakacode/control-plane-flow/issues/372).
+
 ## [5.1.1] - 2026-06-03
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ In addition to the standard keepachangelog.com categories, this project uses a l
 
 ### Fixed
 
-- **Fixed `cpflow` crashing at load time with `invalid byte sequence in US-ASCII (ArgumentError)` on systems without a UTF-8 locale.** `Command::Base.all_commands` now reads command files with an explicit UTF-8 encoding instead of relying on `Encoding.default_external`. Fixes [issue 372](https://github.com/shakacode/control-plane-flow/issues/372).
+- **Fixed `cpflow` crashing at load time with `invalid byte sequence in US-ASCII (ArgumentError)` on systems without a UTF-8 locale.** [PR 404](https://github.com/shakacode/control-plane-flow/pull/404) by [Justin Gordon](https://github.com/justin808). `Command::Base.all_commands` now reads command files with an explicit UTF-8 encoding instead of relying on `Encoding.default_external`. Fixes [issue 372](https://github.com/shakacode/control-plane-flow/issues/372).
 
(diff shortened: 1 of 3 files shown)


justin808:
## Address-review status

Scan scope: initial review feedback on head `b7c9880a5dc3d5cf2b9bba25f98345e1729f0fc8`.

### Mattered

- None.

### Optional

- Changelog PR-link and attribution convention: addressed in e57ca1b by adding the PR 404 link and maintainer attribution.

### Skipped

- None.

The current-head CI and review workflows reran after this commit. Non-cutoff status only; the next review pass must use `check all reviews`.

Issue #375, opened 2026-07-03
CI hygiene: replace deprecated upload-artifact@master, standardize action pinning, add shellcheck gate


justin808:
## Summary

Three related CI/tooling hygiene problems, small individually but worth fixing together (likely as two tiny PRs):

1. `rspec-shared.yml` uses the deprecated, unstable `actions/upload-artifact@master` ref (twice).
2. Action pinning is inconsistent across workflows — `checkout@v3`, `checkout@v6`, and full commit-SHA pins coexist, so there is no single auditable update policy.
3. Shell scripts have no lint gate, and the `.agents/bin/*` wrappers currently trip shellcheck, which blocks ever making shellcheck a CI check.

## Evidence

- [`rspec-shared.yml#L59`](https://github.com/shakacode/control-plane-flow/blob/7c96df930980982160da75d22716202a82a94838/.github/workflows/rspec-shared.yml#L59) and `#L65`: `uses: actions/upload-artifact@master`
- [`rspec-shared.yml#L40`](https://github.com/shakacode/control-plane-flow/blob/7c96df930980982160da75d22716202a82a94838/.github/workflows/rspec-shared.yml#L40), `rubocop.yml#L25`: `actions/checkout@v3` (two majors behind); `claude-code-review.yml` uses `checkout@v6`; `cpflow-cleanup-stale-review-apps.yml` pins checkout by commit SHA
- `shellcheck .agents/bin/*` reports 5× SC1007 for the (deliberate) `CDPATH= cd -- ...` idiom — verified on 7c96df9. Note `script/*` are **Ruby** scripts; the shell surface is `.agents/bin/*`, `lib/generator_templates*/**/*.sh`, `.github/actions/**/*.sh`, `spec/dummy/.controlplane/*.sh`

## Proposed fix

1. `upload-artifact@master` → `@v4` (both occurrences); bump `checkout` to one current major everywhere.
2. Adopt one pinning policy for all workflows. Recommended: major-version tags plus a `dependabot.yml` with the `github-actions` ecosystem so bumps arrive as PRs. (SHA-pinning everywhere is the stricter alternative; either is fine — consistency is the point.)
3. Add a shellcheck CI job scoped to the real shell scripts (list above), fixing the SC1007s via `CDPATH='' cd` or an inline `# shellcheck disable=SC1007` with a comment.

## Acceptance criteria

- [x] No branch refs (`@master`) for any action in `.github/`
[...]

PR #451, opened 2026-09-03
Pin GitHub Actions and generate auditable local actions

justin808:
## Summary

- pin every external GitHub Action to an immutable commit with a reviewed release comment and enforce a closed `trusted_actions` allowlist
- package cpflow's composite actions in the gem and generate/update them under `.github/actions/cpflow-*` alongside downstream workflow wrappers
- make reusable workflows load local actions from the caller repository's trusted event revision while keeping runtime source in the separately pinned `.cpflow` checkout
- add Dependabot coverage, deterministic generator/update contracts, and downstream rollout/testing guidance
- make the generated verification helper fail closed on missing, symlinked, malformed, or path-traversing local action references

Part of #375. PR #446 remains the ShellCheck slice required before closing that issue.

## Trust model

Generated deploy callers check out local actions from the triggering `github.sha`, keeping the wrapper and generated actions synchronized for automatic same-repository pull-request deploys and manually selected refs. Delete callers continue to use the pull request base SHA, falling back to the workflow event SHA outside pull requests, and cleanup uses the event SHA explicitly. Existing downstream installations test an unmerged generated branch with `workflow_dispatch --ref`; first installations validate locally, merge generated files together, and dispatch immediately afterward. Cross-repository reusable-workflow calls remain exact cpflow release tags by design; embedded marketplace actions are full-SHA pinned.

Production dispatch intentionally loads matching workflow and local-action bytes from the selected protected ref. The trust boundary is the `production` Environment with required review, prevent-self-review, and a protected deployable branch/tag; pinning only the action checkout elsewhere would create wrapper/action skew without preventing selected workflow code from accessing the approved environment secret.

## Validation

Exact head: `7068a806c97a2e682b9de7ae3e92b6cb95bc3139`
Integrated base: `9a8992807df2bfc778946350f7643c168c2c33b8`

- focused generator/workflow/policy/docs specs: 153 examples, 0 failures
- RuboCop: 215 files, 0 offenses
- generated command docs: clean
- actionlint: clean (with documented ignores for ShellCheck SC2129 and actionlint's stale reusable-workflow context model)
- exact-head trusted GitHub Actions scanner: 23 files, 0 findings
- gem build/unpack: all 7 canonical composite actions packaged
- `git diff --check`: clean
- CI-parity regression: reproduced with `actionlint` absent, then fixed by isolating the recursive-alias unit spec from that unrelated external tool; the exact selector and full generator spec now pass (1/0 and 83/0)
- review-wrapper skew regression: deploy generated actions now load from the triggering caller revision; focused generator/workflow/docs coverage passes (87/0)
- repeated YAML alias regression: the policy walker terminates recursive aliases but scans every non-recursive alias path; dependency-policy specs pass (17/0)
- downstream custom-action regression: the generated validation helper ignores unrelated local-action namespaces while still rejecting malformed `cpflow-*` paths; combined generator and dependency-policy specs pass (101/0)

The complete credentialed suite was attempted locally but cannot run without `CPLN_ORG`/Control Plane credentials. Exact-head hosted CI remains a mandatory merge gate.


lib/github_flow_templates/.github/workflows/cpflow-promote-staging-to-production.yml
```diff
@@ -171,7 +171,7 @@ jobs:
           } >> "$GITHUB_OUTPUT"
 
       - name: Setup production environment
```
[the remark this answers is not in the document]
justin808: Verified against GitHub dispatch semantics and the full generated workflow. This checkout is not a separate trust boundary: workflow_dispatch already loads the entire caller-owned workflow YAML from the selected ref, so a maliciously selectable ref could read the environment secret directly even if this one checkout were pinned elsewhere. The boundary is the production Environment: required reviewer, prevent self-review, and a protected deployable branch/tag. After that reviewer approves an allowed ref, loading the matching local actions is intentional so the wrapper and generated action contract stay in lockstep; pinning only the action checkout would create version skew without mitigating ref-controlled workflow code. The docs now state the protected production ref requirement consistently. Resolving as not applicable.

.github/workflows/cpflow-delete-review-app.yml
```diff
@@ -563,8 +563,14 @@ jobs:
...
 
+      - name: Checkout generated cpflow actions
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          ref: ${{ github.event.pull_request.base.sha || github.sha }}
+          persist-credentials: false
+
       - name: Checkout control-plane-flow actions
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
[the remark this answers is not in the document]
justin808: Verified at c1b74a155727f908418eb4f1805faf68b8770113 and declining this suggestion as inapplicable. For issue_comment, GitHub loads the downstream wrapper from the default branch and github.sha identifies that triggering commit; workflow_dispatch similarly identifies the selected workflow commit, while PR events supply base.sha. The current base-SHA-or-event-SHA contract therefore keeps generated local actions synchronized with the wrapper that actually ran. Resolving the commented PR base through the API would instead create wrapper/action skew. This contract is documented in CONTRIBUTING.md and docs/ci-automation.md and asserted for both workflows in spec/command/generate_github_actions_spec.rb.

.github/workflows/cpflow-deploy-review-app.yml
```diff
@@ -608,8 +608,14 @@ jobs:
               }
             }
 
+      - name: Checkout generated cpflow actions
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          ref: ${{ github.event.pull_request.base.sha || github.sha }}
```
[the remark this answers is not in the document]
justin808: Verified independently at c1b74a155727f908418eb4f1805faf68b8770113. This duplicates the delete-workflow finding and is declined for the same event-contract reason: issue_comment loads the default-branch wrapper and github.sha points to that triggering commit; workflow_dispatch uses its selected workflow commit; PR events provide base.sha. Loading generated actions from that same revision is intentional and tested. Resolving an arbitrary PR base would break wrapper/action synchronization rather than improve it.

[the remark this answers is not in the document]
justin808: Fixed in 319dc7ec1979b0ae1f73dc7bf949e7c603977427. The deploy workflow now checks out generated local actions from github.sha, so automatic same-repository pull_request runs use the same merge revision as the caller wrapper and manual dispatch uses its selected ref. Delete keeps its base-SHA fallback for pull_request_target safety. Added the event-specific trust model to docs and updated generator coverage; focused specs pass 87/0, actionlint and RuboCop are clean, and the trusted-actions scanner reports 23/0.

lib/github_flow_templates/bin/test-cpflow-github-flow
```diff
@@ -16,10 +16,70 @@ echo "==> parse generated GitHub Actions YAML"
...
+    next if visited_container_ids.key?(node.object_id)
+
+    visited_container_ids[node.object_id] = true
+  end
+
+  case node
+  when Hash
+    uses = node["uses"]
+    if uses.is_a?(String) && uses.start_with?("./.github/actions/")
+      match = uses.match(generated_local_action_pattern)
```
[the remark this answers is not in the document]
justin808: Fixed in 7068a806c97a2e682b9de7ae3e92b6cb95bc3139. The validation helper now reserves strict path validation for the generated ./.github/actions/cpflow-* namespace, so repositories can keep unrelated local actions. The path-traversal regression still uses a cpflow-prefixed reference and remains rejected. Added an end-to-end custom local-action fixture; the combined generator and dependency-policy specs pass 101/0, and RuboCop is clean.

.github/workflows/check_cpln_links.yml
```diff
@@ -6,13 +6,18 @@ on:
...
+  contents: read
+
 jobs:
   check_cpln_links:
     runs-on: ubuntu-latest
     name: Check Links
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
justin808, line 18: why change to sha from version?

.github/workflows/claude-code-review.yml
```diff
@@ -15,13 +17,14 @@ jobs:
...
-        uses: actions/checkout@v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           fetch-depth: 1
+          persist-credentials: false
 
       - name: Run Claude Code Review
         id: claude-review
-        uses: anthropics/claude-code-action@v1
+        uses: anthropics/claude-code-action@8251c103ac8c1d761882c86aba1412c7f583c844 # v1.0.213
```
justin808, line 27: why change to sha from version?

.agents/agent-workflow.yml
```diff
@@ -24,3 +24,16 @@ untrusted_contributor_intake:
   trusted_github_scheme: https
   trusted_github_repo: shakacode/control-plane-flow
 repo_prefix: CPF
+
+# External actions already used by this repository and reviewed for the closed
+# allowlist enforced by the GitHub Actions security scanner. Every use still
+# requires an immutable commit SHA and a same-line release-version comment.
+trusted_actions:
```
justin808, line 31: **Walkthrough 1/5 — Immutable dependencies and least privilege**

Previously, external actions could follow moving major tags/branches, new repositories were not constrained by a closed trust set, and some workflows inherited broad default permissions or all reusable-workflow secrets. This allowlist is the human-reviewed identity boundary; every repository action use is now required to pair a lowercase 40-character SHA with an exact release-tag comment, while Docker actions require a digest. Workflows also declare exact top-level permissions, disable persisted checkout credentials, name reusable-workflow secrets, pin the Control Plane CLI, and scope its token to consuming steps. Dependabot proposes weekly updates, but it cannot silently expand this allowlist.

The main risk is maintenance friction or a stale pin; the security gain is that upstream tag movement and surprise action introduction fail closed. `spec/github_actions_dependency_policy_spec.rb` parses the YAML AST (including hostile alias/merge/multiline forms), reconciles this list, permissions, credentials, secrets, pins, and Dependabot directories; the exact-head scanner reported 23 files with 0 findings.

Does this trust boundary and update burden match what you expect? Reply here, or say **next** for the generation contract.

lib/command/generate_github_actions.rb
```diff
@@ -108,36 +111,57 @@ class GenerateGithubActions < Base
...
 
-    # Resolve template root from __dir__ rather than Cpflow.root_path because this file is
+    # Resolve source roots from __dir__ rather than Cpflow.root_path because this file is
     # loaded before `module Cpflow` finishes defining its class methods.
+    REPOSITORY_ROOT = Pathname.new(File.expand_path("../..", __dir__))
     TEMPLATE_ROOT = Pathname.new(File.expand_path("../github_flow_templates", __dir__))
+    ACTIONS_ROOT = REPOSITORY_ROOT.join(".github/actions")
 
-    def self.generated_files
+    def self.generated_file_sources
```
justin808, line 134: **Walkthrough 2/5 — Package and regenerate local composite actions**

Previously, `generate-github-actions` copied only `lib/github_flow_templates`; downstream repos got workflow wrappers but relied on upstream `.cpflow/.github/actions/*` at runtime. The generator now builds one deterministic source map from both templates and canonical `.github/actions/cpflow-*`, so generation and `update-github-actions` copy workflows, helper scripts, and all seven composite actions together. Existing-file detection covers local actions, updates repair partial/stale trees, executable bits are restored, and snapshot/file-set specs require output to match canonical sources exactly. A gem-packaging spec proves every canonical action is included in the built gem.

The tradeoff is more checked-in generated code downstream, but each local `uses:` target becomes directly auditable and versioned with the wrapper. The key operational risk is wrapper/action skew if maintainers commit only part of an update; later steps show how runtime selection and validation catch that. Focused generator coverage passed as part of the reported 153 examples.

Is the “one gem release produces one complete generated tree” contract clear? Reply here, or say **next** for runtime checkout behavior.

.github/workflows/cpflow-deploy-review-app.yml
```diff
@@ -608,8 +608,14 @@ jobs:
               }
             }
 
+      - name: Checkout generated cpflow actions
```
justin808, line 611: **Walkthrough 3/5 — Load actions from the caller revision; runtime from `.cpflow`**

This new root checkout is the trust-model pivot. The reusable deploy job loads generated `.github/actions/cpflow-*` from the triggering caller revision (`github.sha`) and invokes them as normal local actions, while a second checkout at `.cpflow` remains pinned to GitHub’s `job.workflow_sha` and supplies the cpflow gem/runtime source. That keeps a wrapper and its generated actions synchronized for same-repository PR deploys and manual `workflow_dispatch --ref`, without mixing them with PR application code checked out under `app`. Delete uses the PR base SHA (falling back to the event SHA), cleanup uses the event SHA, and production follows the selected protected ref.

The effect is intentional: staging/review credentials can execute matching workflow plus local-action bytes from the authorized caller revision. The boundary therefore depends on the same-repository caller guard and, for production, the protected `production` Environment, required review/prevent-self-review, and protected deployable refs. A wrong ref could create wrapper/action skew or run untrusted code with credentials; generator specs assert every local action is preceded by a credential-free root checkout with the event-specific ref, while `.cpflow` stays separate.

Does the distinction between caller-owned action code and upstream-pinned runtime source make sense? Reply here, or say **next** for the fail-closed local validator.

lib/github_flow_templates/bin/test-cpflow-github-flow
```diff
@@ -16,10 +16,70 @@ echo "==> parse generated GitHub Actions YAML"
...
+  ".github/workflows/*.yaml"
+].sort.to_h do |path|
+  [path, YAML.load_file(path, aliases: true)]
+end
+
+documents.each_key do |path|
   puts "parsed #{path}"
 end
+
+generated_local_action_pattern = %r{\A\./(?<directory>\.github/actions/cpflow-[a-z0-9]+(?:-[a-z0-9]+)*)\z}
```
justin808, line 32: **Walkthrough 4/5 — Fail closed when generated local actions are unsafe or incomplete**

Once workflows reference checked-in actions, YAML parsing alone is insufficient: a missing directory, symlink, malformed namespace, or traversal such as `cpflow-x/../../…` could make validation bless a different filesystem target. This validator walks all generated workflow/action YAML, accepts only the exact `./.github/actions/cpflow-<slug>` shape, and then uses `lstat` to require a real local directory plus a real `action.yml`/`action.yaml` file. It tracks visited containers so recursive aliases terminate, while unrelated local-action namespaces are deliberately out of cpflow’s ownership.

The validator now fails before `actionlint` or deployment on malformed/missing cpflow refs. The main compatibility risk is rejecting a custom path that pretends to be generated; that is by design. Generator specs cover missing descriptors, traversal, recursive aliases, unrelated actions, and successful canonical trees.

Are these ownership and path-shape limits the right fail-closed behavior? Reply here, or say **next** for the downstream rollout procedure.

justin808 (commented):
## Human walkthrough — COMMENT only

Exact comparison: `main@9a8992807df2bfc778946350f7643c168c2c33b8` → `7068a806c97a2e682b9de7ae3e92b6cb95bc3139` (`jg-codex/375-action-pinning`), 17 commits, 37 files, +1,291/−213.

Purpose: replace moving/unbounded GitHub Actions dependencies with an auditable trust policy, ship cpflow composite actions into downstream repositories, and keep the workflow/action/runtime revisions synchronized. Previously, several dependencies followed tags or branches, reusable workflows invoked composite actions from the separately checked-out upstream source, and downstream generation emitted wrappers but not those action implementations.

This is a **full** five-step walkthrough because the change is both large and security-sensitive:

1. Immutable dependency and least-privilege policy
2. Gem-packaged local-action generation/update contract
3. Caller-revision actions versus pinned runtime source
4. Fail-closed generated-flow validation
5. Downstream rollout and operator guidance

Each conceptual step is a separate inline comment on a representative changed line. Reply in that thread with questions or your checkpoint; walkthrough participation is not approval, review completion, or merge authorization. Current hosted checks shown on the PR are green/skipped, but the PR review decision remains changes-requested and merge readiness is outside this explanatory review.

docs/ci-automation.md
```diff
@@ -852,25 +875,48 @@ releasing it. Use an immutable commit SHA from the upstream PR branch:
    bin/test-cpflow-github-flow ruby /path/to/control-plane-flow/bin/cpflow
    ```
 
-5. Open a downstream PR and trigger the real review app with a comment whose body
-   is exactly:
+5. Push the downstream test branch, then choose the remote validation path that
```
justin808, line 878: **Walkthrough 5/5 — Downstream upgrade and real-world validation path**

The operator contract now says to regenerate and commit wrappers plus `.github/actions/cpflow-*` together. Stable downstream cross-repository reusable-workflow calls intentionally use one exact release tag matching the gem; external actions inside those workflows remain full-SHA pinned. For unreleased testing, regenerate local actions from the upstream checkout, pin all workflow/runtime refs to the upstream head SHA, keep `CPFLOW_VERSION` unset, and run the local contract.

Remote proof differs by installation state: an existing installation can dispatch the unmerged downstream branch with `gh workflow run … --ref`; an initial installation must validate locally, merge the generated files together, then immediately dispatch from the default branch because GitHub will not dispatch a workflow absent from that branch. The docs explicitly reject `+review-app-deploy` for pre-merge wrapper testing because `issue_comment` loads the default-branch workflow. Documentation specs lock these instructions, the exact-release exception, canonical-source locations, and protected production-ref requirement; README, generated command docs, AI prompt, contributor guidance, and changelog carry the same model.

What questions remain about rollout or rollback? If this completes your understanding, say so here; a separate final merge decision prompt will follow only after the discussion is finished.

.agents/agent-workflow.yml
justin808: I think the whole point of using version numbers is to minimize friction. I think version numbers should be used until we detect issues with them. Do we have evidence to suggest that version numbers are not good enough?

justin808: Short answer: **we do not have evidence that a version tag has been exploited or unexpectedly moved in this repository.** This is preventive supply-chain hardening, not a response to a CPF incident.

The evidence for “not good enough” is about the guarantee, not a known failure here:

- GitHub says a full commit SHA is currently the only immutable way to consume an action. Even an exact tag such as `v6.0.2` can be moved or deleted if the upstream repository/account is compromised. GitHub explicitly calls tags more convenient and widely used, but says they retain that risk: https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions
- That consequence matters in this repo because these actions execute inside jobs that may have write-capable `GITHUB_TOKEN` scopes, OIDC, Control Plane credentials, the checked-out workspace, or Docker access. A retargeted tag changes executable code without any CPF diff or review.
- The old state included both moving aliases (`@v1`, `@v3`, `@v4`, `@master`) and exact tags. The PR removes both categories at third-party action boundaries because neither is cryptographically bound to the reviewed commit.

There are two different policies in this PR, intentionally:

1. **Third-party executable actions:** full SHA + same-line exact version comment, e.g. `actions/checkout@de0… # v6.0.2`. The comment preserves human-readable versioning, and Dependabot is configured to propose updates, so the normal upgrade remains a versioned PR rather than manual SHA archaeology.
2. **Our own cpflow reusable workflows:** exact release tags such as `v5.3.0` remain allowed. That is the deliberate convenience exception because we control the tag/release process and need one readable version to align the gem, generated wrappers/actions, and upstream workflow. Moving aliases such as `v5` are still not allowed.

So the honest tradeoff is: exact tags reduce friction, while SHAs prevent a tag rewrite from silently changing code. There is no local incident proving we *must* choose SHAs; the current choice follows GitHub’s strongest published guidance for high-impact third-party code, and uses Dependabot plus release comments to recover most of the maintenance ergonomics.

Does that evidence justify the preventive boundary for third-party actions, or would you prefer that this PR explicitly accept tag-rewrite risk and use exact version tags there too? I’ll keep this thread open either way; this is a policy decision, not an approval checkpoint.

